PR #3, opened 2026-06-13
Fix run_tool dropping empty arguments for no-arg downstream tools

eshwar-sundar-glean:
## Problem

`run_tool` failed for every downstream tool called with no arguments (e.g. `slack_read_user_profile`), returning:

> Expected: object, given: null

## Root cause

In `src/tools/run-tool.ts`, the payload assembly omitted the `arguments` key entirely when the resolved args were empty:

```js
if (Object.keys(resolvedArgs).length > 0) {
  remoteArgs.arguments = resolvedArgs;
}
```

With the key absent, the backend `run_tool` meta-tool forwarded a downstream `tools/call` with no `arguments`, which serializes to `null`. Strict downstream MCP servers validate `params.arguments` as an object and reject `null`.

The key's **presence**, not its value, was the only difference between a failing and succeeding call — passing any non-empty object (the prior workaround) incidentally tripped the guard true.

## Fix

Always include `arguments`, defaulting to an explicit `{}` when empty — matching what the MCP SDK sends for direct tool calls. Extracted the assembly into a pure `buildRemoteArgs()` helper and added regression tests for the empty and populated cases.

## Verification

- `npm run typecheck` — clean
- `npm test` — 134 passed, including 2 new `buildRemoteArgs` tests

### DIFF
--- a/packages/codex/dist/index.js
+++ b/packages/codex/dist/index.js
@@ -31152,14 +31152,18 @@ async function handleRunTool(remoteClient, mcpServer, skillsBaseDir, args) {
     }
     throw err;
   }
-  const remoteArgs = {
+  return callRemoteTool(
+    remoteClient,
+    "run_tool",
+    buildRemoteArgs(serverId, toolName, resolvedArgs)
+  );
+}
+function buildRemoteArgs(serverId, toolName, resolvedArgs) {
+  return {
     server_id: serverId,
-    tool_name: toolName
+    tool_name: toolName,
+    arguments: resolvedArgs
   };
-  if (Object.keys(resolvedArgs).length > 0) {
-    remoteArgs.arguments = resolvedArgs;
-  }
-  return callRemoteTool(remoteClient, "run_tool", remoteArgs);
 }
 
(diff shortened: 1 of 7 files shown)


plugins/glean/.claude-plugin/plugin.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "glean-experimental",
-  "version": "0.2.17",
+  "version": "0.2.18",
```
mohit-gupta-glean, line 3: #suggestion, in a separate PR, can we add a check to make sure that the 3 versions remain in sync.

packages/codex/dist/index.js
```diff
@@ -31152,14 +31152,18 @@ async function handleRunTool(remoteClient, mcpServer, skillsBaseDir, args) {
     }
     throw err;
   }
-  const remoteArgs = {
+  return callRemoteTool(
```
mohit-gupta-glean, line 31155: Add this file to gitignore.

mohit-gupta-glean (approved):

PR #6, opened 2026-06-15
[Plugins] Readable, viewport-safe HITL approval prompt (compact args + spill file)

swarup-padhi-glean:
## Problem

When `ENABLE_HITL` is on, `run_tool` shows a `requires_approval` action's arguments in an `elicitInput` prompt. Two issues with the original rendering:

1. **Unreadable args** — the whole object was rendered with `JSON.stringify(args, null, 2)`, which escapes newlines inside string values, so a Markdown body/table collapsed into one line of literal `\n`.
2. **Viewport overflow** — once arguments were expanded for readability, a large argument (long body, table, etc.) could make the prompt taller than Claude Code's viewport and push the **Accept/Decline buttons off-screen**, so the user couldn't respond.

## Fix

The inline approval message is now **compact and line-capped** so the buttons always stay visible, with full detail spilled to a file when needed.

**Inline message (Claude Code / Codex):**
- One line per top-level argument; **UPPERCASE keys**, indented under `Arguments:` — distinguishes structure from values in plain text (the elicitation surface can't style fonts).
- Each value on a single line: multi-line strings collapsed; values past ~120 chars cut and suffixed `(truncated)`; nested objects shown as compact JSON.
- Capped to a small argument-section budget (≈8 lines — up to 7 args + 1 file-path line) so Accept/Decline stay in view.

**Spill file (only when something is truncated or omitted):**
- Full arguments are written to a single fixed Markdown file in the plugin's data dir (`PLUGIN_DATA_DIR`), overwritten on each approval; the prompt shows the path.
- Rendered for readability when opened: string values verbatim (so Markdown — tables, headings — renders), nested values as pretty JSON.
- The write is wrapped in try/catch — if a write is blocked (e.g. a restrictive sandbox), it degrades to a short note instead of breaking the approval gate.

**Cursor** keeps its one-line prompt (it shows tool + args in its own UI).

Example inline prompt (fabricated values):
```
Action: send_message
Arguments:
  CHANNEL: C123
  BODY: # Notes | Feature | Status | |---|---| | Search | done | ... (truncated)
  Full arguments: <PLUGIN_DATA_DIR>/glean-approval-args.md
```

## Scope & non-goals
- **Plain text** — the elicitation surface doesn't render Markdown; the inline ceiling is readable text, while the spill file is where rich content renders.
- Affects only the Claude Code / Codex message; Cursor unchanged.
- Complements `file_args`, which keeps huge inputs out of `arguments` entirely.

## Code
- Argument-formatting helpers extracted into **`src/tools/approval-args.ts`** (`buildCompactArgs`, `formatArgumentsForFile`, `writeApprovalArgsFile`); `run-tool.ts` builds the message and gates the file write.

## Tests
- Unit tests for the compact formatter (scalars inline, multi-line collapse, over-width truncation with `(truncated)`, nested compaction, line-budget cap) and the file formatter; the HITL handler test asserts the prompt stays within the line budget and the spill file holds the full content. Build + typecheck + tests all green.

## Verification
- Verified in Claude Code, including **`/sandbox` mode**: the full-args file writes successfully to the plugin's managed data dir (the same location the plugin already uses for its own state), so the sandbox permits it; the try/catch covers the edge case where a write is blocked.


.claude/settings.json
```diff
@@ -1,5 +1,3 @@
 {
-  "enabledPlugins": {
-    "glean-vnext@glean-plugins-vnext": true
-  }
+  "enabledPlugins": {}
```
eshwar-sundar-glean: Why? We should probably add extraKnownMarketplaces if you are seeing an issue with this

src/tools/run-tool.ts
```diff
@@ -141,33 +142,168 @@ function isCursorClient(mcpServer: Server): boolean {
...
+// faithfully inline (multi-line, long, or extra arguments) is written in full
+// to a file and the path is shown instead — never expanded inline.
+// The argument section of the approval prompt is capped to this many lines so
+// the Accept/Decline buttons stay in view. When a spill file is needed, one of
+// these lines is the file path (so up to maxArgSectionLines-1 arguments show).
+const maxArgSectionLines = 8;
+// Per-argument inline width before truncating with an ellipsis.
+const maxApprovalArgChars = 120;
+
+function safeJson(value: unknown): string {
```
eshwar-sundar-glean: #suggest
add a utils file and move all the utils there

src/tools/run-tool.ts
```diff
@@ -141,33 +142,168 @@ function isCursorClient(mcpServer: Server): boolean {
...
   try {
-    return JSON.stringify(args, null, 2);
+    return JSON.stringify(value);
   } catch {
-    return String(args);
+    return String(value);
   }
 }
 
+function isEmptyArgs(args: unknown): boolean {
```
eshwar-sundar-glean: same as above, utils file

src/tools/run-tool.ts
```diff
@@ -141,33 +142,168 @@ function isCursorClient(mcpServer: Server): boolean {
...
+      truncated = true;
+    }
+    rendered =
+      collapsed.length > maxApprovalArgChars
+        ? `${collapsed.slice(0, maxApprovalArgChars)}…`
+        : collapsed;
+  } else if (value !== null && typeof value === "object") {
+    const json = safeJson(value);
+    if (json.length > maxApprovalArgChars) {
+      rendered = `${json.slice(0, maxApprovalArgChars)}…`;
```
eshwar-sundar-glean: maybe suffix with (truncated)?

src/tools/run-tool.ts
```diff
@@ -141,33 +142,168 @@ function isCursorClient(mcpServer: Server): boolean {
...
+  // labels (Action / Arguments) stay distinct from the values; keys are
+  // uppercased (in compactArgLine) so a key reads distinctly from its value.
+  // Both are plain-text cues that spend no vertical space.
+  const message = [
+    `Action: ${toolName}`,
+    "Arguments:",
+    ...lines.map((line) => `  ${line}`),
+  ];
+  if (needsFile) {
+    const filePath = await writeApprovalArgsFile(toolName, args);
```
eshwar-sundar-glean: #blocking
Do test this out.
We should try to run this in /sandbox mode in claude, we are writing files outside project dir

eshwar-sundar-glean (approved):

swarup-padhi-glean:
Tested in claude code CLI:
Screenshot 2026-06-20 at 10 29 25 AM

Tested in claude code CLI sandbox
Screenshot 2026-06-20 at 10 34 54 AM

Arguments:
Screenshot 2026-06-20 at 10 36 30 AM

cc: @eshwar-sundar-glean

eshwar-sundar-glean (approved):
LGTM

src/tools/approval-args.ts
```diff
@@ -0,0 +1,133 @@
...
+    const collapsed = value.replace(/\s+/g, " ").trim();
+    if (value.includes("\n") || collapsed.length > maxApprovalArgChars) {
+      truncated = true;
+    }
+    rendered =
+      collapsed.length > maxApprovalArgChars
+        ? `${collapsed.slice(0, maxApprovalArgChars)}… (truncated)`
+        : collapsed;
+  } else if (value !== null && typeof value === "object") {
+    const json = safeJson(value);
```
mohit-gupta-glean, line 51: #clarify Are we not handling nested jsons? Should we have a recursive printer at least in the file if not here.

src/tools/approval-args.ts
```diff
@@ -0,0 +1,133 @@
...
+    } else {
+      out.push("```json", JSON.stringify(value, null, 2), "```", "");
+    }
+  }
+  return out.join("\n");
+}
+
+// The full arguments are written to a single fixed file under the plugin's
+// data dir (CLAUDE_PLUGIN_DATA, exported by start.sh as PLUGIN_DATA_DIR). It is
+// intentionally overwritten on each approval — only the most recent prompt's
```
mohit-gupta-glean: #suggest what if there are multiple sessions active in parallel. This may be rare but will lead to overwrites. We should do one per session.

mohit-gupta-glean (approved):

src/tools/approval-args.ts
```diff
@@ -0,0 +1,133 @@
...
+export async function writeApprovalArgsFile(
+  toolName: string,
+  args: unknown,
+): Promise<string> {
+  const dir =
+    process.env.PLUGIN_DATA_DIR ||
+    process.env.CLAUDE_PLUGIN_DATA ||
+    os.tmpdir();
+  await fs.mkdir(dir, { recursive: true });
+  const file = path.join(dir, "glean-approval-args.md");
```
mohit-gupta-glean, line 134: #suggest can we make the path tmp/{Session_id}/

src/tools/approval-args.ts
swarup-padhi-glean: Nested JSON is handled — `JSON.stringify` is recursive, so a separate printer isn't needed:

- **Inline prompt:** nested objects/arrays render as compact one-line JSON via `JSON.stringify(value)`, then truncated to the inline width like any value.
- **Spill file:** nested values are pretty-printed via `JSON.stringify(value, null, 2)` inside a fenced code block — fully expanded at any depth.

One nuance worth flagging: only **top-level** string values are written verbatim in the file (so Markdown like tables renders). A string nested *inside* an object value appears within that object's JSON block (escaped, not Markdown-rendered). If we want nested Markdown strings to render too, that's a small follow-up (walk the object and emit nested string values as their own sections) — happy to add it if you think it's worth it.

PR #9, opened 2026-06-17
[NI][Plugins] Derive Glean Server URL from work email in setup

pragati-agrawal-glean:
setup now accepts a work email and resolves the Server (QE) URL via the public config/search endpoint, instead of requiring the admin-only about-glean URL. Falls back to manual server_url entry when the domain isn't recognized; multi-tenant deployments resolve correctly.

Bumps esbuild->0.28.1 and pins vite->7.3.5 (overrides) to clear CVE blocks that prevented the build from installing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,19 @@ Today it ships one plugin:
 
 ## First run
 
-The first tool call triggers OAuth sign-in to Glean via the `setup` tool.
-The agent's skill walks the user through the sign-in URL and paste-back
-flow. After sign-in, OAuth credentials are cached to `~/.glean/` and reused
-across sessions — you won't be prompted again until the refresh token
-expires.
+Setup resolves your Glean instance automatically from your work email, then
+triggers OAuth sign-in to Glean via the `setup` tool. 
+
+### Setting the Server URL manually
+
+You normally don't need this — `setup` derives the Server (QE) URL from your
+work email. Set it explicitly only when you're testing a specific deployment
+or your email domain isn't recognized. Two options:
+
+- Pass `server_url` to the `setup` tool with your **Server instance (QE)** URL
+  (e.g. `https://acme-be.glean.com`). Admins can find this at
+  https://app.glean.com/admin/about-glean.
+- Or set the `GLEAN_MCP_SERVER_URL` environment variable.
 
 ## Updates
 
--- a/plugins/glean/skills/glean_run/SKILL.md
+++ b/plugins/glean/skills/glean_run/SKILL.md
@@ -22,8 +22,8 @@ returns a response containing `[SETUP_REQUIRED]`, the user needs to
 When this happens:
 1. Call `setup` (no arguments).
    - If no Server URL is configured, `setup` returns `[SETUP_REQUIRED]` with
-     instructions. Relay them, ask the user for their Glean Server instance
-     (QE) URL, then call `setup` again with `server_url` set to that URL.
+     instructions. Relay them, ask the user for their work email, then call
+     `setup` again with `email` set to what they provided. 
    - Once a Server URL is configured, `setup` opens the Glean sign-in page in
      the browser and waits for sign-in.
 2. Once `setup` returns "Glean setup is complete", retry the original tool
--- a/src/config-search.ts
+++ b/src/config-search.ts
@@ -1,0 +1,92 @@
+const CONFIG_SEARCH_URL = "https://app.glean.com/config/search";
+
+// Shape of the deployment config returned by the config/search endpoint.
+interface DeploymentConfig {
+  queryURL?: string;
+  centralURL?: string;
+  isMultiTenant?: boolean;
+}
+
+interface ConfigSearchResponse {
+  search_config?: DeploymentConfig;
+}
+
+export type ResolveResult =
+  | { ok: true; queryUrl: string }
+  | { ok: false; error: string };
+
+const emailPattern = /^[^@\s]+@([^@\s]+\.[^@\s]+)$/;
+
+export async function resolveServerUrlFromEmail(
+  email: string,
+  fetchImpl: typeof fetch = fetch,
+): Promise<ResolveResult> {
+  const trimmed = email.trim();
+  const match = emailPattern.exec(trimmed);
+  if (!match) {
+    return { ok: false, error: `"${email}" is not a valid email address.` };
+  }
+  const emailDomain = match[1].toLowerCase();
+
+  let resp: Response;
+  try {
+    resp = await fetchImpl(CONFIG_SEARCH_URL, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Accept: "application/json",
+      },
+      body: JSON.stringify({ email: trimmed, emailDomain, isGleanApp: true }),
+    });
+  } catch (err) {
+    const msg = err instanceof Error ? err.message : String(err);
+    return {
+      ok: false,
+      error: `Could not reach Glean to look up your instance: ${msg}`,
+    };
+  }
+
+  if (!resp.ok) {
+    return {
+      ok: false,
+      error: `Glean instance lookup failed (HTTP ${resp.status}).`,
+    };
+  }
+
+  let data: ConfigSearchResponse;
+  try {
+    data = (await resp.json()) as ConfigSearchResponse;
+  } catch {
+    return {
+      ok: false,
+      error: "Glean instance lookup returned an unexpected response.",
+    };
+  }
+
+  const cfg = data.search_config;
+  const queryUrl = cfg?.queryURL;
+
+  // No usable instance if there's no queryURL, or the endpoint returned the
+  // shared central URL for an unknown domain or typo in email
+  if (
+    !queryUrl ||
+    (cfg?.centralURL &&
+      cfg.isMultiTenant !== true &&
+      sameOrigin(queryUrl, cfg.centralURL))
+  ) {
+    return {
+      ok: false,
+      error: `No Glean instance is registered for "${emailDomain}".`,
+    };
+  }
+
+  return { ok: true, queryUrl };
+}
+
+function sameOrigin(a: string, b: string): boolean {
+  try {
+    return new URL(a).origin === new URL(b).origin;
+  } catch {
+    return false;
+  }
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,6 +40,7 @@ import {
   type DispatchContext,
 } from "./tools/remote-passthrough.js";
 import { resolveSessionId } from "./session-id.js";
+import { resolveServerUrlFromEmail } from "./config-search.js";
 
 function readEnv(...keys: string[]): string | undefined {
   for (const key of keys) {
@@ -64,11 +65,14 @@ function normalizeServerUrl(raw: string): string {
 
 const SETUP_REQUIRED_TEXT =
   `[SETUP_REQUIRED]\n\n` +
-  `To connect this plugin to your Glean instance:\n` +
-  `1. Visit https://app.glean.com/admin/about-glean (log in if needed)\n` +
-  `2. Copy the **Server instance (QE)** URL shown on that page (e.g. https://acme-be.glean.com/)\n` +
-  `3. Paste it here\n\n` +
-  `Then call this tool again with the server_url parameter set to the URL you copied.`;
+  `To connect, enter your work email (e.g. you@acme.com) and we'll find ` +
+  `your Glean instance automatically.\n`;
+
+// A failed lookup asks the user to retry with a corrected email. 
+// Admins who must set the URL manually use
+// the server_url parameter (documented in the README)
+const EMAIL_RESOLVE_FAILED_TEXT =
+  `Double-check the email for typos and try again with the corrected email.`;
 
 const SETUP_NEEDED_ERROR =
   "Glean is not configured yet. Call the `setup` tool first to provide " +
@@ -232,18 +236,27 @@ const SETUP_TOOL: Tool = {
   annotations: { readOnlyHint: true },
   description:
     "Check or configure the Glean connection. Setup completes in three " +
-    "stages: (1) save the Server URL, (2) sign in, (3) fetch the remote " +
-    "tool catalog. Call with no arguments to advance through the next " +
-    "missing stage; it opens the Glean sign-in page in the browser when " +
-    "needed. Call with server_url to (re)configure. Call with reset=true to " +
-    "clear all configuration.",
+    "stages: (1) resolve and save the Server URL, (2) authenticate, " +
+    "(3) fetch the remote tool catalog. Call with no arguments to advance " +
+    "through the next missing stage. Call with email to look up and " +
+    "(re)configure user's Glean instance. Call with reset=true to clear " +
+    "all configuration.",
   inputSchema: {
     type: "object" as const,
     properties: {
+      email: {
+        type: "string",
+        description:
+          "User's work email (e.g. you@acme.com). Used to look up and " +
+          "configure your Glean Server instance (QE) URL automatically.",
+      },
       server_url: {
         type: "string",
         description:
-          "Glean Server Instance (QE) URL (e.g. https://acme-be.glean.com).",
+          "Advanced. Sets the Glean Server (QE) URL directly instead of " +
+          "resolving it from email. Do not suggest this to users — email " +
+          "is the preferred path. Documented in the README for technical " +
+          "users who ask for it explicitly.",
       },
       reset: {
         type: "boolean",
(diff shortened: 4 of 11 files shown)


src/index.ts
```diff
@@ -246,18 +254,30 @@ const SETUP_TOOL: Tool = {
...
-    "next missing stage. Call with server_url to (re)configure. Call with " +
-    "callback_url to finish authentication after a sign-in paste. Call " +
-    "with reset=true to clear all configuration.",
+    "stages: (1) resolve and save the Server URL, (2) authenticate, " +
+    "(3) fetch the remote tool catalog. Call with no arguments to advance " +
+    "through the next missing stage. Call with email to look up and " +
+    "configure your Glean instance automatically (the normal path). Call " +
+    "with server_url to set the Server URL directly (fallback when the " +
+    "email domain isn't recognized). Call with callback_url to finish " +
+    "authentication after a sign-in paste. Call with reset=true to clear " +
```
eshwar-sundar-glean: Can you test once with a dumber model like Sonnet with medium thinking to make sure it follows all these cases:
1. Email + Auth + Reset
2. Wrong email + wrong server url + correct server url + auth + reset
3. Wrong email + correct server url + auth + reset

src/index.ts
```diff
@@ -246,18 +254,30 @@ const SETUP_TOOL: Tool = {
...
+    "all configuration.",
   inputSchema: {
     type: "object" as const,
     properties: {
+      email: {
+        type: "string",
+        description:
+          "Your work email (e.g. you@acme.com). Used to look up your " +
```
eshwar-sundar-glean: nit User's work email?

src/index.ts
```diff
@@ -665,8 +706,7 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
                 type: "text",
                 text:
                   `Invalid URL: "${rawUrl}". Please provide the Server instance (QE) URL ` +
-                  `from https://app.glean.com/admin/about-glean ` +
-                  `(e.g. https://acme-be.glean.com).`,
+                  `(e.g. https://acme-be.glean.com), or pass your email instead.`,
```
eshwar-sundar-glean, line 751: #clarify
any reason to remove this line `from https://app.glean.com/admin/about-glean`

src/config-search.ts
```diff
@@ -0,0 +1,104 @@
+// Resolves a user's Glean Server (QE) URL from their work email via the public
+// config/search endpoint — the same lookup the web app uses to route users to
+// their tenant. Lets `setup` derive the URL instead of asking for it manually.
+
+const CONFIG_SEARCH_URL = "https://app.glean.com/config/search";
+
+// Mirrors com/askscio config_endpoint DeploymentConfig.
```
eshwar-sundar-glean: #suggest
We may / may not accept contributions, its good to keep the repo self contained to public code and not reference implementation details / names from internal code base

src/config-search.ts
```diff
@@ -0,0 +1,104 @@
...
+      ok: false,
+      error: `No Glean instance is registered for "${emailDomain}".`,
+    };
+  }
+
+  // Unknown domains aren't an error here — the endpoint returns the shared
+  // central URL with isMultiTenant=false. Reject that so a typo doesn't
+  // silently point at the central endpoint (the user can still use
+  // server_url). Multi-tenant customers legitimately have queryURL ===
+  // centralURL but with isMultiTenant=true, so accept those.
```
eshwar-sundar-glean: clarify, what do you mean by this
Reject that so a typo doesn't
  // silently point at the central endpoint (the user can still use
  // server_url)

packages/codex/dist/index.js
mohit-gupta-glean, line 1: rebase to latest master. The codex file is gone.

src/config-search.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  let resp: Response;
+  try {
+    resp = await fetchImpl(CONFIG_SEARCH_URL, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Accept: "application/json",
+      },
+      body: JSON.stringify({ email: trimmed, emailDomain, isGleanApp: true }),
```
mohit-gupta-glean, line 39: #clarify what is isGleanApp?

src/index.ts
```diff
@@ -62,11 +63,18 @@ function normalizeServerUrl(raw: string): string {
 
 const SETUP_REQUIRED_TEXT =
   `[SETUP_REQUIRED]\n\n` +
-  `To connect this plugin to your Glean instance:\n` +
+  `To connect, either:\n` +
+  `- enter your work email (e.g. you@acme.com) and we'll find your Glean instance automatically, or\n` +
+  `- enter your Glean Server (QE) URL directly, if you have it.\n\n` +
```
mohit-gupta-glean: #suggest can we not talk about the URL flow if email works. That means that only show this if email does not work for some reason.

src/index.ts
```diff
@@ -62,11 +63,18 @@ function normalizeServerUrl(raw: string): string {
...
+  `Then call this tool again with the email or server_url parameter set.`;
+
+// Shown when an email's domain can't be resolved to a Glean instance. Falls
+// back to the original manual method.
+const MANUAL_URL_FALLBACK_TEXT =
```
mohit-gupta-glean: #suggest Let's also add an option to try again with valid email to correct for typos.

src/index.ts
```diff
@@ -246,18 +254,30 @@ const SETUP_TOOL: Tool = {
...
     "Check or configure the Glean connection. Setup completes in three " +
-    "stages: (1) save the Server URL, (2) authenticate, (3) fetch the " +
-    "remote tool catalog. Call with no arguments to advance through the " +
-    "next missing stage. Call with server_url to (re)configure. Call with " +
-    "callback_url to finish authentication after a sign-in paste. Call " +
-    "with reset=true to clear all configuration.",
+    "stages: (1) resolve and save the Server URL, (2) authenticate, " +
+    "(3) fetch the remote tool catalog. Call with no arguments to advance " +
+    "through the next missing stage. Call with email to look up and " +
+    "configure your Glean instance automatically (the normal path). Call " +
```
mohit-gupta-glean: nit: preferred

src/index.ts
```diff
@@ -246,18 +254,30 @@ const SETUP_TOOL: Tool = {
...
+    "with server_url to set the Server URL directly (fallback when the " +
```
mohit-gupta-glean: nit: not needed? There may be other reasons to set server URL directly

src/index.ts
```diff
@@ -246,18 +254,30 @@ const SETUP_TOOL: Tool = {
...
+    "authentication after a sign-in paste. Call with reset=true to clear " +
+    "all configuration.",
   inputSchema: {
     type: "object" as const,
     properties: {
+      email: {
+        type: "string",
+        description:
+          "Your work email (e.g. you@acme.com). Used to look up your " +
+          "Glean Server instance (QE) URL automatically. Preferred over " +
```
mohit-gupta-glean: This is very confusing. Here, we are saying preferred over server_url and then in server_url, we are saying "takes precedence over email".

mohit-gupta-glean (commented):
#suggest Can we get rid of the fallback flow and simplify. We should document the server URL flag instead (in README, not in user journey). No point in complicating the setup and the fallback path anyway won't work for majority of users as they are not admins.

src/index.ts
pragati-agrawal-glean: done
Thanks

src/index.ts
pragati-agrawal-glean: done
thanks

src/index.ts
pragati-agrawal-glean: done
thanks

src/index.ts
pragati-agrawal-glean: It is removed because this link was admin-only, so the two available options are 1. to provide the qe-url directly (if they have) or give their email. The link is shown if they reach the fallback path (line #75). As per https://github.com/gleanwork/glean-plugins-vnext/pull/9#pullrequestreview-4529913844, would show it only when email path does not work

src/config-search.ts
pragati-agrawal-glean: So for customers whose isMultiTenant is true, it means they do not have a dedicated URL for them and are still on queryURL = centralURL = https://apps-be.glean.com/ . The comment here says that if someone does a typo in the email domain, that should not connect it to the centralURL but creates an error.

src/config-search.ts
pragati-agrawal-glean: It is a parameter required in the query posted to app.glean.com as mentioned [here](https://docs.google.com/document/d/122kbwGc9R1WaKTS2GD3OJCbyuQkw0XXyBeLRaV7eczg/edit?tab=t.0). It's a flag indicating the caller is a first-party Glean app/client (as opposed to some embedded/third-party context).

Screenshot 2026-06-19 at 12 16 48 PM

eshwar-sundar-glean:
There are multiple merge conflicts, can we rebase?

src/config-search.ts
pragati-agrawal-glean: done
thanks

packages/codex/dist/index.js
pragati-agrawal-glean: done

src/index.ts
pragati-agrawal-glean: #clarify Removed the URL flow completely as it is accessible only by admins

src/index.ts
pragati-agrawal-glean: done
thanks

src/index.ts
pragati-agrawal-glean: https://github.com/gleanwork/glean-plugins-vnext/pull/9#discussion_r3447381352

src/config-search.ts
```diff
@@ -0,0 +1,95 @@
+// Resolves a user's Glean Server (QE) URL from their work email via the public
+// config/search endpoint
```
swarup-padhi-glean: #nit: We can do away with this comment as this does give something about our systems.

plugins/glean/skills/glean_run/SKILL.md
```diff
@@ -34,9 +35,14 @@ other way.
...
-returns instructions — follow them and call `setup` again; it guides the whole
-flow. Once it returns "Glean setup is complete", proceed to Step 1.
+Call `setup` (with no arguments) to confirm the Glean connection is
+configured.
+
+- If it returns `[SETUP_REQUIRED]`, relay the instructions to the user and
+  wait for them to provide their work email. Then call `setup` again with
+  `email` set to what they provided. If the lookup fails, ask for the email
+  again — a failed lookup is almost always a typo.
+- Once `setup` confirms configuration is complete, proceed to Step 1.
```
swarup-padhi-glean: #clarify: Was there a particular reason to enhance this, seems the original step 0 description was good enough.

src/config-search.ts
pragati-agrawal-glean: done
thanks

plugins/glean/skills/glean_run/SKILL.md
pragati-agrawal-glean: tried to make it easier for the model to understand. 
would revert back to original.
thanks

src/index.ts
pragati-agrawal-glean: the flow is much simpler now with the server url being removed from description but being in the readme.

src/config-search.ts
```diff
@@ -0,0 +1,92 @@
...
+  | { ok: false; error: string };
+
+const emailPattern = /^[^@\s]+@([^@\s]+\.[^@\s]+)$/;
+
+export async function resolveServerUrlFromEmail(
+  email: string,
+  fetchImpl: typeof fetch = fetch,
+): Promise<ResolveResult> {
+  const trimmed = email.trim();
+  const match = emailPattern.exec(trimmed);
```
swarup-padhi-glean, line 25: nit: inline trimmed

src/config-search.ts
```diff
@@ -0,0 +1,92 @@
...
+  const cfg = data.search_config;
+  const queryUrl = cfg?.queryURL;
+
+  // No usable instance if there's no queryURL, or the endpoint returned the
+  // shared central URL for an unknown domain or typo in email
+  if (
+    !queryUrl ||
+    (cfg?.centralURL &&
+      cfg.isMultiTenant !== true &&
+      sameOrigin(queryUrl, cfg.centralURL))
```
swarup-padhi-glean, line 75: #clarify: What is centralUrl and when is it returned?

src/index.ts
```diff
@@ -64,11 +65,16 @@ function normalizeServerUrl(raw: string): string {
...
   `[SETUP_REQUIRED]\n\n` +
-  `To connect this plugin to your Glean instance:\n` +
-  `1. Visit https://app.glean.com/admin/about-glean (log in if needed)\n` +
-  `2. Copy the **Server instance (QE)** URL shown on that page (e.g. https://acme-be.glean.com/)\n` +
-  `3. Paste it here\n\n` +
-  `Then call this tool again with the server_url parameter set to the URL you copied.`;
+  `To connect, enter your work email (e.g. you@acme.com) and we'll find ` +
+  `your Glean instance automatically.\n` +
+  `Then call this tool again with the email parameter set.`;
+
```
swarup-padhi-glean: #clarify: is this an instruction to the LLM or to the user. If later then ```Then call this tool again with the email parameter set``` seems weird. If LLM then this should be in third person in my opinion.

src/index.ts
```diff
@@ -64,11 +65,16 @@ function normalizeServerUrl(raw: string): string {
...
+// A failed lookup is almost always a typo, so the journey just asks the user
+// to retry with a corrected email. Admins who must set the URL manually use
+// the server_url parameter (documented in the README); we don't walk through
+// it here since it needs admin access that most users don't have.
+const EMAIL_RESOLVE_FAILED_TEXT =
+  `Double-check the email for typos and call setup again with the corrected email.`;
```
swarup-padhi-glean: #clarify: same as above

swarup-padhi-glean (approved):
LGTM. Wait for either @mohit-gupta-glean or @eshwar-sundar-glean before merge.

src/config-search.ts
pragati-agrawal-glean: centralURL is https://apps-be.glean.com/ - a shared/central endpoint for customers who are on the multi-tenant deployment and don't have their queryURL configured yet.

src/index.ts
eshwar-sundar-glean: +1 enter user's email would be good

plugins/glean/skills/glean_run/SKILL.md
```diff
@@ -22,8 +22,9 @@ returns a response containing `[SETUP_REQUIRED]`, the user needs to
 When this happens:
 1. Call `setup` (no arguments).
    - If no Server URL is configured, `setup` returns `[SETUP_REQUIRED]` with
-     instructions. Relay them, ask the user for their Glean Server instance
-     (QE) URL, then call `setup` again with `server_url` set to that URL.
+     instructions. Relay them, ask the user for their work email, then call
+     `setup` again with `email` set to what they provided. If the lookup
```
eshwar-sundar-glean:
```
If the lookup
     fails, ask for the email again — a failed lookup is almost always a typo.
```
This is not required since you are returning this text from the response too

src/config-search.ts
eshwar-sundar-glean: So even if we get a query URL but there is a central url set and the customer is single tenant and the query url and centralURL are same origin we throw. I dont understand why?

so does it mean this endpoint sometimes return apps-be url even for queryUrl? that looks like a bug on the endpoint

src/index.ts
pragati-agrawal-glean: It was for the user. Removed that line from the instruction:
const SETUP_REQUIRED_TEXT =
  `[SETUP_REQUIRED]\n\n` +
  `To connect, enter your work email (e.g. you@acme.com) and we'll find ` +
  `your Glean instance automatically.\n`;

src/index.ts
```diff
@@ -232,18 +238,27 @@ const SETUP_TOOL: Tool = {
...
+      },
       server_url: {
         type: "string",
         description:
-          "Glean Server Instance (QE) URL (e.g. https://acme-be.glean.com).",
+          "Advanced. Sets the Glean Server (QE) URL directly instead of " +
+          "resolving it from email. Do not suggest this to users — email " +
+          "is the standard path. Documented in the README for technical " +
+          "users who ask for it explicitly.",
       },
```
eshwar-sundar-glean, line 260: Are we keeping this param, technical users might just set the env variable?

README.md
```diff
@@ -32,11 +32,22 @@ Today it ships one plugin:
 
 ## First run
 
-The first tool call triggers OAuth sign-in to Glean via the `setup` tool.
-The agent's skill walks the user through the sign-in URL and paste-back
-flow. After sign-in, OAuth credentials are cached to `~/.glean/` and reused
-across sessions — you won't be prompted again until the refresh token
-expires.
+Setup resolves your Glean instance automatically from your work email, then
+triggers OAuth sign-in to Glean via the `setup` tool. The agent's skill walks
```
eshwar-sundar-glean:
```
The agent's skill walks
the user through the email prompt, the sign-in URL, and the paste-back flow.
After sign-in, OAuth credentials are cached to `~/.glean/` and reused across
sessions — you won't be prompted again until the refresh token expires.
```

This entire text is not needed

eshwar-sundar-glean (approved):
LGTM on logic, try to address other comments before merge.

src/index.ts
pragati-agrawal-glean: For user. 

const EMAIL_RESOLVE_FAILED_TEXT =
  `Double-check the email for typos and try again with the corrected email.`;

src/index.ts
pragati-agrawal-glean: This is for the LLM to infer if someone wants to configure using a link from Readme. Can remove if unnecessary. Please suggest

README.md
pragati-agrawal-glean: done
Thanks

src/index.ts
eshwar-sundar-glean: right yes, fair point, i have seen LLMs read the readme in cursor and codex

PR #8, opened 2026-06-17
Simplify Codex Plugin after update. Align plugin versions.

eshwar-sundar-glean:
## Summary
- Move the Codex plugin into `plugins/glean` (`.codex-plugin` manifest + assets) so all three harnesses live under one plugin directory.
- Align all plugin manifests to **0.2.23**, fixing version drift between harnesses.

## Version alignment
The manifests had drifted out of sync:

| Manifest | Before | After |
|---|---|---|
| `.claude-plugin/plugin.json` | 0.2.21 | **0.2.23** |
| `.cursor-plugin/plugin.json` | 0.2.19 | **0.2.23** |
| `.codex-plugin/plugin.json` | 0.2.19 | **0.2.23** |

All three now share a single incremented version.

PR #11, opened 2026-06-18
Mark dist bundles as generated to suppress diffs

mohit-gupta-glean:
## Summary
- Add `packages/codex/dist/**` to `.gitattributes` with `linguist-generated` and `-diff` so GitHub collapses them in PR review and `git diff` suppresses their content
- Update the pre-commit rebuild hook to use an ignorelist — skip rebuild only when ALL staged files are non-source (docs, CI config, git metadata, hooks). New source paths automatically trigger rebuilds without needing to update the hook.

## Test plan
- [x] `git diff --exit-code` still detects out-of-sync dist (exits non-zero with "Binary files differ")
- [x] CI `check-version-bump` unaffected — `.gitattributes` and `.githooks/` don't match `PLUGIN_PATHS`
- [x] Commits touching only ignored paths (e.g. `.gitattributes`) skip rebuild and version bump

### DIFF
--- a/.githooks/pre-commit.d/01-rebuild-dist.sh
+++ b/.githooks/pre-commit.d/01-rebuild-dist.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
-# Rebuild dist/index.js and stage it on every commit.
+# Rebuild dist/index.js and stage it on every commit that touches source files.
 set -e
 cd "$(git rev-parse --show-toplevel)"
+
+# Skip rebuild if only non-source files are staged (docs, CI, git metadata).
+IGNORE_PATHS="^(\.git(attributes|ignore|hooks/)|\.github/|README|LICENSE|CHANGELOG|\.node-version)"
+if ! git diff --cached --name-only | grep -qvE "$IGNORE_PATHS"; then
+  exit 0
+fi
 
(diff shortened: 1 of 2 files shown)


.githooks/pre-commit.d/01-rebuild-dist.sh
```diff
@@ -1,8 +1,14 @@
 #!/bin/bash
-# Rebuild dist/index.js and stage it on every commit.
+# Rebuild dist/index.js and stage it on every commit that touches source files.
```
eshwar-sundar-glean, line 2: https://github.com/gleanwork/glean-plugins-vnext/pull/8

Probably not needed, Codex seems to be fixed now

mohit-gupta-glean: Merging this anyway. You can drop it as part of your PR. Don't think having the attribute will break anything.

mohit-gupta-glean: There are also other useful changes in the PR.

eshwar-sundar-glean: Yea, codex as well does not honour in its new update

eshwar-sundar-glean (approved):

PR #12, opened 2026-06-19
Use CURSOR_CONVERSATION_ID as session id on Cursor (with temp debug logs)

swarup-padhi-glean:
## What
Maps Cursor's `CURSOR_CONVERSATION_ID` to the normalized `GLEAN_SESSION_ID` in `plugins/glean/start.sh`, mirroring how `CLAUDE_CODE_SESSION_ID` is handled. Host-awareness stays in the launcher; the bundle keeps reading only `GLEAN_SESSION_ID` (`src/session-id.ts` unchanged in behavior). Cursor previously fell back to a per-process UUID; now it uses the real conversation id.

## ⚠️ Temporary debug logs — REMOVE BEFORE MERGE
Marked `DEBUG (remove before merge)`:
- `plugins/glean/start.sh`: stderr line showing which host var fed the session id.
- `src/session-id.ts`: stderr line showing the resolved id (host vs fallback).

These are only to verify the wiring in Cursor; strip them (and rebuild `dist`) before merge.

## Version
Manifests bumped 0.2.24 → 0.2.25.

## Note
The old `CODEX_THREAD_ID` mapping was dropped when `packages/codex` was removed; this PR scopes to Cursor only.

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.24",
+  "version": "0.2.25",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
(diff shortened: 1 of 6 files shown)


swarup-padhi-glean:
@mohit-gupta-glean @eshwar-sundar-glean unfortunately cursor doesn't inject the CURSOR_CONVERSATION_ID var into the mcp server process env, so this isn't avl and we will have to fallback to random uuid.

PR #13, opened 2026-06-19
Promote chat, memory, user_activity, employee_search to first-class tools

eshwar-sundar-glean:
## Summary

Two related changes:

### 1. Promote five Glean remote tools to first-class local tools

Promotes the following alongside the existing `search` and `read_document`:

- `chat`
- `memory`
- `memory_schema`
- `user_activity`
- `employee_search`

`chat` was previously excluded because the upstream Glean MCP backend kept
returning `"response contains message of type ERROR"`. That backend issue is
resolved, so `chat` is now re-enabled.

### 2. Extend the version-bump CI check to all plugin manifests

The plugin ships one manifest per host (Claude, Codex, Cursor), but the
version-bump check only validated the Claude one — so the Codex and Cursor
manifests had silently drifted (both `0.2.24` while Claude moved to `0.2.25`).

`scripts/check-version-bump.sh` now validates all three manifests: each must
be valid semver, all must share one aligned version, and that version must be
strictly greater than the highest version on the base branch. The change
trigger also fires on the codex/cursor manifests and `.mcp.codex.json`. The
Codex and Cursor manifests are aligned to `0.2.25`.

## Changes

- `src/tools/remote-passthrough.ts` — add the five tools to `REMOTE_TOOLS_ALLOWLIST`.
- `src/tools/descriptions.ts` — extend the shared-description doc comment to cover the newly promoted tools.
- `tests/remote-passthrough.test.ts` — update the allowlist assertion to the full promoted set; switch the "excluded tool" fixture off `chat`.
- `scripts/check-version-bump.sh` — validate all three manifests (aligned + bumped + no downgrade); broaden the change trigger.
- `plugins/glean/.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`, `.cursor-plugin/plugin.json` — bump to `0.2.25` (aligned).
- `plugins/glean/dist/index.js` — rebuilt bundle (esbuild 0.28.1, matching the lockfile so CI's rebuild matches byte-for-byte).
- `.gitignore` — ignore the local `.claude/tmp/` skills cache.

## Testing

- `npm run test` — 145/145 passing.
- `bash scripts/check-version-bump.sh origin/main` — verified pass (aligned + bumped), and verified it fails on drift and on a non-bumped version.

### DIFF
--- a/src/tools/remote-passthrough.ts
+++ b/src/tools/remote-passthrough.ts
@@ -7,13 +7,17 @@ import {
   type RemoteClientOptions,
 } from "../remote-client.js";
 
-// `chat` is intentionally excluded for now: the upstream Glean MCP backend
-// keeps returning "Error running chat tool: response contains message of
-// type ERROR" with no structured detail, so surfacing it would expose a
-// broken first-class tool. Re-add once the backend is healthy.
+// Remote tools promoted to first-class local tools once setup completes.
+// Anything the remote MCP server exposes that is not in this set is dropped
+// by fetchAllowedRemoteTools and rejected by dispatchRemoteTool.
 export const REMOTE_TOOLS_ALLOWLIST: ReadonlySet<string> = new Set([
   "search",
   "read_document",
+  "chat",
+  "memory",
+  "memory_schema",
+  "user_activity",
+  "employee_search",
 ]);
 
 type ToolInputSchema = Tool["inputSchema"];
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 /dist/
 .DS_Store
+.claude/tmp/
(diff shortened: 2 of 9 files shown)


swarup-padhi-glean (approved):

src/tools/remote-passthrough.ts
```diff
@@ -7,13 +7,17 @@ import {
...
 } from "../remote-client.js";
 
-// `chat` is intentionally excluded for now: the upstream Glean MCP backend
-// keeps returning "Error running chat tool: response contains message of
-// type ERROR" with no structured detail, so surfacing it would expose a
-// broken first-class tool. Re-add once the backend is healthy.
+// Remote tools promoted to first-class local tools once setup completes.
+// Anything the remote MCP server exposes that is not in this set is dropped
+// by fetchAllowedRemoteTools and rejected by dispatchRemoteTool.
 export const REMOTE_TOOLS_ALLOWLIST: ReadonlySet<string> = new Set([
```
mohit-gupta-glean, line 13: @eshwar-sundar-glean do we need a second allowlist here? Can't we rely on what the remote server is returning?

PR #14, opened 2026-06-19
Replace OAuth paste-back with in-context loopback sign-in

eshwar-sundar-glean:
## Summary

Replaces the OAuth **paste-back** sign-in with an **in-context loopback** flow.
Because the plugin runs on the same machine as the browser, OAuth can complete
on a local loopback server instead of routing the code through a hosted "Copy
URL → paste into chat" step.

`setup` now opens the Glean sign-in page in the browser, and a loopback server
at `http://127.0.0.1:29107/glean-cli-callback` captures the authorization code
automatically. No copy/paste, no `callback_url`.

## What changed

- **Removed the paste-back flow entirely** — `callback_url` param,
  `applyPastedCallbackUrl`, `extractAuthCode`, the hosted callback page, the
  `[AUTHENTICATION_REQUIRED]` paste prose, and the cross-process
  `pending-auth-store`.
- **Fixed loopback port.** Glean **exact-matches** the DCR-registered
  `redirect_uri` (no wildcard port), so DCR, the authorize request, and the
  token exchange must all use the identical URL. The loopback binds a fixed port
  (`29107`, overridable via `GLEAN_CALLBACK_PORT`) at `/glean-cli-callback`.
- **No URL-mode elicitation.** Sign-in opens the browser directly; the loopback
  captures the code, then `finishAuth` exchanges it for tokens.
- **Diagnostics.** `setup` logs the negotiated client capabilities.
- Updated `SKILL.md` auth section; bumped all three plugin manifests to
  `0.2.26` (aligned).

## Behavioral notes

- **In-context only.** Auth now requires the browser and plugin on the same
  machine (the loopback must be reachable). The hosted-callback path that
  supported remote execution is gone.
- **DCR is unchanged in spirit** — the plugin still registers dynamically (no
  static client_id) and reuses the persisted client. Recovery via
  `needsFreshClient()` and `setup({reset:true})` is retained.

## Testing

- `npm test` — 136/136 passing (callback-server tests bind an ephemeral port
  via `GLEAN_CALLBACK_PORT=0` to avoid collisions).
- `npm run build` — dist rebuilt with esbuild 0.28.1 and verified in sync.
- `bash scripts/check-version-bump.sh origin/main` — `0.2.25 → 0.2.26 (all
  manifests aligned)`.

Rebased onto current `main` (post #13).


plugins/glean/skills/glean_run/SKILL.md
```diff
@@ -16,43 +16,33 @@ listing available skills.
...
-Do not pass `callback_url` to any tool other than `setup` — only `setup`
-accepts it. Do not treat `[SETUP_REQUIRED]` or `[AUTHENTICATION_REQUIRED]`
-as an error or attempt to work around it any other way.
+1. Call `setup` (no arguments).
+   - If no Server URL is configured, `setup` returns `[SETUP_REQUIRED]` with
+     instructions. Relay them, ask the user for their Glean Server instance
+     (QE) URL, then call `setup` again with `server_url` set to that URL.
+   - Once a Server URL is configured, `setup` opens the Glean sign-in page in
+     the browser and waits for sign-in. The authorization code is captured
+     automatically on a local loopback — there is **nothing to paste**.
```
eshwar-sundar-glean:
```
The authorization code is captured
     automatically on a local loopback — there is **nothing to paste**.
```


This is not needed

src/auth-callback-server.ts
```diff
@@ -1,68 +1,126 @@
...
+      const returnedState = reqUrl.searchParams.get("state");
+      if (returnedState !== expectedState) {
+        res.writeHead(403, { "Content-Type": "text/html" });
+        res.end(
+          "<html><body><h1>Error</h1><p>Invalid OAuth state parameter.</p></body></html>",
+        );
+        rejectCode(new Error("OAuth state mismatch — possible CSRF attack"));
         return;
       }
+    }
```
eshwar-sundar-glean, line 81: this is being set on the Oauth server callback now, so honoring it

src/index.ts
```diff
@@ -328,103 +310,163 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
+      authUrl = err.authUrl;
+    }
+
+    // Open the Glean sign-in page; the loopback (already listening) captures
+    // the redirect automatically once the user signs in.
+    openBrowser(authUrl);
+
+    let code: string;
+    try {
+      code = await withTimeout(handle.code, SIGN_IN_WAIT_MS);
```
eshwar-sundar-glean, line 421: this is why ellicitation is not required

plugins/glean/skills/glean_run/SKILL.md
```diff
@@ -16,43 +16,33 @@ listing available skills.
...
 
 Call `setup` (with no arguments) to confirm the Glean connection is
-configured.
+configured and signed in.
 
-- If it returns `[SETUP_REQUIRED]`, relay the instructions to the user and
-  wait for them to paste their Server URL. Then call `setup` again with
-  `server_url` set to what they pasted.
-- Once `setup` confirms configuration is complete, proceed to Step 1.
+- If it returns `[SETUP_REQUIRED]`, relay it and call `setup` again with
```
mohit-gupta-glean: remove the URL param mention. We should just ask to call setup again. Setup should guide the flow.

plugins/glean/skills/glean_run/SKILL.md
```diff
@@ -16,43 +16,33 @@ listing available skills.
...
+  `server_url` set to the URL the user provides.
+- Otherwise `setup` opens the browser for sign-in when needed and captures
```
mohit-gupta-glean: This also seems somewhat redundant.

mohit-gupta-glean (approved):

src/index.ts
```diff
@@ -328,103 +310,163 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
-      text:
-        "No pending authentication flow found. The callback_url parameter " +
-        "is only valid after a prior call returned [AUTHENTICATION_REQUIRED]. " +
-        "Please initiate the tool call without callback_url first.",
-      isError: true,
-    };
+// How long to wait for the user to complete the browser sign-in (open the
+// authorize URL, sign in, redirect to the loopback) before giving up. The
+// loopback capture is the real wait; this bounds the blocking setup call.
+const SIGN_IN_WAIT_MS = 300_000;
```
mohit-gupta-glean, line 316: How long does the host wait before timing out? In early days, I had noticed timeouts at 60s.

mohit-gupta-glean (approved):
#clarify Can you share a screenshot of the auth completion landing page? Can we add some messaging there to go back to chat and say auth is done.

mohit-gupta-glean (commented):
## AI-Assisted Review

**Overall: Well-structured simplification.** The PR cleanly replaces a multi-step paste-back flow with an in-context loopback server. The deletion of `pending-auth-store.ts`, `descriptions.ts`, and all paste-back machinery is a meaningful UX win. Code is clear and well-tested.

### Issues

**1. Module-level mutable state in `auth-callback-server.ts` — no cleanup on unexpected process exit**

`active` and `expectedState` are module-level singletons. If the process crashes or is killed between `startCallbackServer()` and `closeCallbackServer()`, the port stays bound until the OS reaps it. This is likely fine in practice (port 29107 will be freed), but worth noting that there's no `process.on('exit')` or `process.on('SIGTERM')` guard.

**2. `connectWithSignIn` calls `createRemoteClient` up to 3 times**

The function attempts connect (fast path with existing tokens), then connect again (to trigger AuthRequiredError and get the authorize URL), then a third time after getting the code. The second call is just to extract `authUrl` from the error — would it be cleaner to have the provider expose the authorize URL directly rather than round-tripping through the full client creation? Not a bug, but adds latency (DCR + network roundtrip) on the sign-in path.

**3. `closeCallbackServer()` resets `expectedState` — potential race**

In `closeCallbackServer()`, setting `expectedState = undefined` clears state validation. If a stale browser redirect arrives between when the code is resolved and when `close` is called (e.g. double-click on the IdP), the server would accept it with no state check. This window is tiny and the server is being torn down anyway, so low risk.

**4. The `finally { closeCallbackServer() }` in `connectWithSignIn` closes the server even on success**

This means if the same `setup` call is invoked again quickly (e.g. retry logic), a new server will be created. That's fine given the idempotent `startCallbackServer`, but worth confirming that's intentional.

**5. `conn.client!` non-null assertion in `advanceSetup`**

The types allow `{ client?: RemoteClient; result?: CallToolResult }` where both could be undefined. The assertion is safe given the `if (conn.result) return` guard above, but a discriminated union (`{ kind: 'ok', client } | { kind: 'err', result }`) would eliminate the assertion.

### Nits

- The `dist/index.js` diff is 90% variable renames (`path8→path7`, `fs8→fs7`) from esbuild's scope hoisting after removing the pending-auth module. Expected but noisy.
- Good use of `GLEAN_CALLBACK_PORT=0` in tests for ephemeral binding.

### Summary

Ship-ready with the minor suggestions above. The non-null assertion (5) is the most actionable change; the rest are observations.

plugins/glean/skills/glean_run/SKILL.md
eshwar-sundar-glean: done

src/index.ts
eshwar-sundar-glean: tried for 3 minutes min

eshwar-sundar-glean:
@mohit-gupta-glean addressed some of the AI assisted reviews, not others which did not seem important

PR #15, opened 2026-06-19
Update README file to match the current state of the Glean Plugin

eshwar-sundar-glean:
The Glean Plugin README was not upto date in terms of capabilities and setup, this PR fixes it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Glean plugin
 
 Glean's plugin for [Claude Code](https://code.claude.com/docs/en/overview),
-[Claude Cowork](https://claude.com/cowork), Codex, and Cursor. This repo is a
-single-plugin marketplace, shipping one manifest per host.
+Codex, and Cursor. This repo is a single-plugin marketplace, shipping one
+manifest per host.
 
 Today it ships one plugin:
 
@@ -23,13 +23,6 @@ Today it ships one plugin:
 /plugin marketplace add gleanwork/glean-plugins-vnext
 /plugin install glean-vnext@glean-plugins-vnext
 ```
-
-### Claude Cowork (desktop)
-
-1. Open the plugin picker.
-2. Click **Add marketplace**, choose **GitHub**, and enter
-   `gleanwork/glean-plugins-vnext`.
-3. Once the marketplace syncs, install the **glean-vnext** plugin from it.
 
 ### Codex and Cursor
 
@@ -67,11 +60,7 @@ or your email domain isn't recognized. Two options:
 ## Updates
 
 ```
-# Claude Code
 /plugin marketplace update glean-plugins-vnext
-
-# Cowork: the plugin picker has a "Sync" / "Check for updates"
-# button on the marketplace entry.
 ```
 
 ## Testing a specific branch or PR
@@ -191,7 +180,7 @@ runtime lives under `plugins/glean/`. See the Layout section below.
 
 ```
 .claude-plugin/
-  marketplace.json        Marketplace manifest for Claude Code / Cowork.
+  marketplace.json        Marketplace manifest for Claude Code.
                           Points at ./plugins/glean as the plugin source.
 .cursor-plugin/
   marketplace.json        Marketplace manifest for Cursor.
--- a/README.md
+++ b/README.md
@@ -102,11 +102,14 @@ Then just `git checkout` whichever branch you want to test.
 
 ## Configuration
 
-The plugin is configured entirely through environment variables. None are
-required for normal use — the launcher (`plugins/glean/start.sh`) derives the
-storage and session variables from what the host provides, and the server URL
-is captured by the `setup` tool on first run. The variables below are read
-directly by the server bundle.
+Primary configuration is interactive, not environment-based: the `setup` tool
+captures your Glean Server URL and drives OAuth sign-in on first run (see
+[First run](#first-run)), persisting both under `~/.glean/`. The environment
+variables below are optional and read directly by the server bundle —
+`GLEAN_MCP_SERVER_URL` overrides the URL saved by `setup`, the storage and
+session variables are normally derived by the launcher
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
 
 | Variable | Purpose | Default |
 | --- | --- | --- |
--- a/README.md
+++ b/README.md
@@ -123,11 +123,11 @@ variables the bundle reads, so these are internal — start.sh overwrites them o
 every launch and setting them yourself has no effect:
 
 - `PLUGIN_DATA_DIR` — directory for credentials, caches, the saved server URL,
-  `glean-server.log`, and (planned) staged write-tool arguments. Defaults to
-  `~/.glean`, or the host's `CLAUDE_PLUGIN_DATA` dir when provided.
-- `SKILLS_BASE_DIR` — where discovered skill files are written; anchored under
-  the host's `CLAUDE_PLUGIN_DATA` dir (or `~/.claude/tmp`), or redirected under
-  the launch project when `USE_CLAUDE_PROJECT_DIR=1`.
+  and `glean-server.log`. Defaults to `~/.glean`, or the host's
+  `CLAUDE_PLUGIN_DATA` dir when provided.
+- `SKILLS_BASE_DIR` — where discovered skill files are written; defaults to
+  `<PLUGIN_DATA_DIR>/glean-skills-cache`, or redirected under the launch
+  project's `.claude/tmp/` when `USE_CLAUDE_PROJECT_DIR=1`.
 - `GLEAN_SESSION_ID` — the host's conversation id: `CLAUDE_CODE_SESSION_ID` for
   Claude Code, `CODEX_THREAD_ID` for Codex, otherwise a generated UUID.
 
--- a/README.md
+++ b/README.md
@@ -26,20 +26,11 @@ Today it ships one plugin:
 
 ### Codex and Cursor
 
-The same plugin ships host-specific manifests so it can be installed from
-Codex and Cursor as well:
-
-- **Codex** reads `.agents/plugins/marketplace.json` (marketplace
-  `glean-plugins-vnext`) and the per-host manifest at
-  `plugins/glean/.codex-plugin/plugin.json`, which points the MCP server at
-  `.mcp.codex.json`.
-- **Cursor** reads `.cursor-plugin/marketplace.json` (marketplace
-  `glean-plugins`) and `plugins/glean/.cursor-plugin/plugin.json`, which reuses
-  the shared `.mcp.json`.
-
-Point the host at this repo (`gleanwork/glean-plugins-vnext`) and install
-**glean-vnext** through that host's plugin flow. The launcher, skills, and
-server bundle are shared across all hosts.
+This repo is also a plugin marketplace for **Codex** and **Cursor**: point the
+host at `gleanwork/glean-plugins-vnext` and install **glean-vnext** through that
+host's plugin flow. The launcher, skills, and server bundle are shared across
+all hosts. For Cursor, see
+[Team marketplaces](https://cursor.com/docs/plugins#team-marketplaces).
 
 ## First run
 
@@ -108,50 +99,43 @@ Then just `git checkout` whichever branch you want to test.
 
 ## Configuration
 
-Primary configuration is interactive, not environment-based: the `setup` tool
-captures your Glean Server URL and drives OAuth sign-in on first run (see
-[First run](#first-run)), persisting both under `~/.glean/`. The environment
-variables below are optional and read directly by the server bundle —
-`GLEAN_MCP_SERVER_URL` overrides the URL saved by `setup`, the storage and
-session variables are normally derived by the launcher
-(`plugins/glean/start.sh`) from what the host provides, and the rest tune
-runtime behavior. None are required for normal use.
+Configuration is interactive: the `setup` tool captures your Glean Server URL
+and drives OAuth sign-in on first run (see [First run](#first-run)), persisting
+both under `~/.glean/`. No environment variables are required.
+
+A few optional variables let you override behavior. Set them in the host's MCP
+server `env` block — the shipped `.mcp.json` / `.mcp.codex.json` already set the
+HITL ones — or in your shell:
 
 | Variable | Purpose | Default |
 | --- | --- | --- |
-| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL. When unset, the URL saved by the `setup` tool is used. | (stored config from `setup`) |
-| `ENABLE_HITL` | Enables human-in-the-loop confirmation before `run_tool` executes a downstream tool. Active only when set to exactly `true`. | disabled |
-| `HITL_TIMEOUT_MS` | Timeout, in milliseconds, for a human-in-the-loop confirmation prompt. Must be a positive integer. | `300000` (5 min) |
-| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size, in bytes, of each file read via `run_tool`'s `file_args`. Must be a positive integer. | `1048576` (1 MiB) |
-| `PLUGIN_DATA_DIR` | Directory for cached credentials, the remote-tools cache, the saved server URL, and `glean-server.log`. Will also hold staged arguments for write tools. | `~/.glean` |
-| `SKILLS_BASE_DIR` | Directory where discovered skill files are written. | `/tmp/glean-skills-cache` |
-| `GLEAN_SESSION_ID` | Chat session id sent with backend calls. | a UUID generated once per process |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL captured by `setup`. | URL saved by `setup` |
+| `ENABLE_HITL` | Human-in-the-loop confirmation before `run_tool` runs a downstream tool. Active only when set to exactly `true`. | `true` (set in the shipped `.mcp.json`) |
+| `HITL_TIMEOUT_MS` | Timeout in milliseconds for a HITL confirmation prompt. Positive integer. | `300000` (5 min), set in `.mcp.json` |
+| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size in bytes of each file read via `run_tool`'s `file_args`. Positive integer. | `1048576` (1 MiB), bundle default |
+| `USE_CLAUDE_PROJECT_DIR` | Set to `1` to route the skills cache under the launch project's `.claude/tmp/`, so the `glean_run` skill's `Read` glob can match cache files. | unset |
 
-Empty values and un-interpolated `${VAR}` placeholders are ignored, so a host
-that passes an unset variable through verbatim falls back to the default.
+Empty values and un-interpolated `${VAR}` placeholders are ignored, falling back
+to the default.
 
-The defaults above are the bundle's built-in fallbacks. The shipped MCP
-manifests (`plugins/glean/.mcp.json` and `.mcp.codex.json`) set
-`ENABLE_HITL=true` and `HITL_TIMEOUT_MS=300000` in their `env` block, so
-human-in-the-loop confirmation is **on by default** in the packaged plugin.
+The launcher (`plugins/glean/start.sh`) also derives and **exports** three more
+variables the bundle reads, so these are internal — start.sh overwrites them on
+every launch and setting them yourself has no effect:
 
-### Launcher-managed variables
-
-`start.sh` runs host-side and normalizes the following host inputs into the
-plugin variables above, keeping the server bundle host-agnostic:
-
-| Host variable | Effect |
-| --- | --- |
-| `CLAUDE_PLUGIN_DATA` | Host-managed lifecycle dir. When set, both `PLUGIN_DATA_DIR` and `SKILLS_BASE_DIR` (as `<dir>/glean-skills-cache`) are anchored under it. |
-| `USE_CLAUDE_PROJECT_DIR` | Opt-in (`=1`): routes the skills cache under the launch project's `.claude/tmp/` so the `glean_run` skill's `Read` glob can match cache files. |
-| `CLAUDE_CODE_SESSION_ID` | Claude Code's conversation id; copied into `GLEAN_SESSION_ID` so the session id tracks the host conversation. |
-| `CODEX_THREAD_ID` | Codex's conversation id; used as `GLEAN_SESSION_ID` when `CLAUDE_CODE_SESSION_ID` is unset. (Cursor exposes no session id, so the plugin generates one.) |
-| `HOME` | Fallback base for `PLUGIN_DATA_DIR` (`~/.glean`) and `SKILLS_BASE_DIR` when `CLAUDE_PLUGIN_DATA` is absent. |
+- `PLUGIN_DATA_DIR` — directory for credentials, caches, the saved server URL,
+  `glean-server.log`, and (planned) staged write-tool arguments. Defaults to
+  `~/.glean`, or the host's `CLAUDE_PLUGIN_DATA` dir when provided.
+- `SKILLS_BASE_DIR` — where discovered skill files are written; anchored under
+  the host's `CLAUDE_PLUGIN_DATA` dir (or `~/.claude/tmp`), or redirected under
+  the launch project when `USE_CLAUDE_PROJECT_DIR=1`.
+- `GLEAN_SESSION_ID` — the host's conversation id: `CLAUDE_CODE_SESSION_ID` for
+  Claude Code, `CODEX_THREAD_ID` for Codex, otherwise a generated UUID.
 
 ## Troubleshooting
 
-- **Sign-in loop** — the cached OAuth provider state may be stale. Delete
-  `~/.glean/mcp-credentials.json` and retry.
+- **Sign-in loop or stale auth** — call the `setup` tool with `reset=true` to
+  clear the saved configuration and credentials, then call `setup` again to
+  re-authenticate.
 - **Tools return `[SETUP_REQUIRED]`** — the plugin isn't configured or
   authenticated yet. Call the `setup` tool (no arguments) to advance through
   the next missing stage: saving the Server URL, signing in, then fetching the
--- a/README.md
+++ b/README.md
@@ -1,18 +1,19 @@
-# Glean Claude plugin
+# Glean plugin
 
-Glean's plugin for [Claude Code](https://code.claude.com/docs/en/overview)
-and [Claude Cowork](https://claude.com/cowork). This repo is a single-plugin
-marketplace.
+Glean's plugin for [Claude Code](https://code.claude.com/docs/en/overview),
+[Claude Cowork](https://claude.com/cowork), Codex, and Cursor. This repo is a
+single-plugin marketplace, shipping one manifest per host.
 
 Today it ships one plugin:
 
-- **`glean-vnext`** — adds two static tools, `find_skills` and `run_tool`, that
-  let the agent discover Glean-hosted skills for enterprise apps (Jira, Slack,
-  Google Workspace, Salesforce, etc.) and invoke their downstream tools via
-  Glean's MCP gateway. Once the user has authenticated, the plugin also
-  surfaces a small allow-list of Glean's first-class tools — currently
-  `search` and `read_document` — directly, with schemas pulled from the
-  remote MCP server.
+- **`glean-vnext`** — adds three static tools, `find_skills`, `run_tool`, and
+  `setup`, that let the agent discover Glean-hosted skills for enterprise apps
+  (Jira, Slack, Google Workspace, Salesforce, etc.) and invoke their downstream
+  tools via Glean's MCP gateway. Once the user has authenticated, the plugin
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -85,12 +92,52 @@ For local development, point the marketplace at your local checkout instead:
 
 Then just `git checkout` whichever branch you want to test.
 
+## Configuration
+
+The plugin is configured entirely through environment variables. None are
```
eshwar-sundar-glean: The plugin is configured entirely through environment variables. --> this is not true, this is rather the fallback for stuff like GLEAN_MCP_SERVER_URL. The setup is done interactively calling /glean_run setup

eshwar-sundar-glean: Good catch — fixed in 8fde9d1. Reworded the opening so it leads with the interactive `setup` tool capturing the Server URL + OAuth, and frames the env vars as optional overrides/fallbacks (`GLEAN_MCP_SERVER_URL` overriding the setup-saved URL) plus launcher-derived storage/session vars and tuning knobs.

README.md
```diff
@@ -1,18 +1,19 @@
-# Glean Claude plugin
+# Glean plugin
 
-Glean's plugin for [Claude Code](https://code.claude.com/docs/en/overview)
-and [Claude Cowork](https://claude.com/cowork). This repo is a single-plugin
-marketplace.
+Glean's plugin for [Claude Code](https://code.claude.com/docs/en/overview),
```
swarup-padhi-glean, line 3: Do we need this doc mention in the Readme?  I don't think is needed but my suggestion is if we do want to have it, we might also want the same for codex and cursor.

README.md
```diff
@@ -85,12 +92,55 @@ For local development, point the marketplace at your local checkout instead:
...
+`GLEAN_MCP_SERVER_URL` overrides the URL saved by `setup`, the storage and
+session variables are normally derived by the launcher
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
+
+| Variable | Purpose | Default |
+| --- | --- | --- |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL. When unset, the URL saved by the `setup` tool is used. | (stored config from `setup`) |
+| `ENABLE_HITL` | Enables human-in-the-loop confirmation before `run_tool` executes a downstream tool. Active only when set to exactly `true`. | disabled |
+| `HITL_TIMEOUT_MS` | Timeout, in milliseconds, for a human-in-the-loop confirmation prompt. Must be a positive integer. | `300000` (5 min) |
```
swarup-padhi-glean: Can we also mention this is activated only when ENABLE_HITL is set to true

README.md
```diff
@@ -41,11 +52,7 @@ expires.
...
 
 ```
-# Claude Code
 /plugin marketplace update glean-plugins-vnext
-
-# Cowork: the plugin picker has a "Sync" / "Check for updates"
-# button on the marketplace entry.
 ```
 
 ## Testing a specific branch or PR
```
swarup-padhi-glean, line 61: When talking about testing, do we not also want to do it for other surfaces like cursor and codex. Here I have for claude code and codex https://docs.google.com/document/d/1mn6v0gmIzaWerdPfZGumnaaCM01wJ8gZnwEKFNssjQY/edit?tab=t.0.

README.md
```diff
@@ -23,12 +24,22 @@ Today it ships one plugin:
 /plugin install glean-vnext@glean-plugins-vnext
 ```
 
-### Claude Cowork (desktop)
```
swarup-padhi-glean, line 26: Do we not want to mention installation steps for claude code desktop at least?

README.md
```diff
@@ -85,12 +92,55 @@ For local development, point the marketplace at your local checkout instead:
...
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
+
+| Variable | Purpose | Default |
+| --- | --- | --- |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL. When unset, the URL saved by the `setup` tool is used. | (stored config from `setup`) |
+| `ENABLE_HITL` | Enables human-in-the-loop confirmation before `run_tool` executes a downstream tool. Active only when set to exactly `true`. | disabled |
+| `HITL_TIMEOUT_MS` | Timeout, in milliseconds, for a human-in-the-loop confirmation prompt. Must be a positive integer. | `300000` (5 min) |
+| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size, in bytes, of each file read via `run_tool`'s `file_args`. Must be a positive integer. | `1048576` (1 MiB) |
+| `PLUGIN_DATA_DIR` | Directory for cached credentials, pending-auth state, the remote-tools cache, the saved server URL, and `glean-server.log`. | `~/.glean` |
```
swarup-padhi-glean: We would also be using this for arguments storage for write tools.

README.md
```diff
@@ -85,12 +92,55 @@ For local development, point the marketplace at your local checkout instead:
...
+| `SKILLS_BASE_DIR` | Directory where discovered skill files are written. | `/tmp/glean-skills-cache` |
+| `GLEAN_SESSION_ID` | Chat session id sent with backend calls. | a UUID generated once per process |
+
+Empty values and un-interpolated `${VAR}` placeholders are ignored, so a host
+that passes an unset variable through verbatim falls back to the default.
+
+The defaults above are the bundle's built-in fallbacks. The shipped MCP
+manifests (`plugins/glean/.mcp.json` and `.mcp.codex.json`) set
+`ENABLE_HITL=true` and `HITL_TIMEOUT_MS=300000` in their `env` block, so
```
swarup-padhi-glean: Hey I think we would want to increase the HITL_TIMEOUT. 30000 ms is just 0.5 minutes. @eshwar-sundar-glean can you bump this up to 5 minutes I guess? Corresponding this doc will need a change.

README.md
```diff
@@ -85,12 +92,55 @@ For local development, point the marketplace at your local checkout instead:
...
+`start.sh` runs host-side and normalizes the following host inputs into the
+plugin variables above, keeping the server bundle host-agnostic:
+
+| Host variable | Effect |
+| --- | --- |
+| `CLAUDE_PLUGIN_DATA` | Host-managed lifecycle dir. When set, both `PLUGIN_DATA_DIR` and `SKILLS_BASE_DIR` (as `<dir>/glean-skills-cache`) are anchored under it. |
+| `USE_CLAUDE_PROJECT_DIR` | Opt-in (`=1`): routes the skills cache under the launch project's `.claude/tmp/` so the `glean_run` skill's `Read` glob can match cache files. |
+| `CLAUDE_CODE_SESSION_ID` | Copied into `GLEAN_SESSION_ID` so the session id tracks the host conversation. |
+| `HOME` | Fallback base for `PLUGIN_DATA_DIR` (`~/.glean`) and `SKILLS_BASE_DIR` when `CLAUDE_PLUGIN_DATA` is absent. |
+
```
swarup-padhi-glean, line 131: We missed moving the CODEX_THREAD_ID var as per this comment https://github.com/gleanwork/glean-plugins-vnext/pull/8/changes#r3445450021.

Don't think it's a beta blocker but we can either add in code and here or do nothing for now.

README.md
eshwar-sundar-glean: No

README.md
eshwar-sundar-glean: will do now

README.md
eshwar-sundar-glean: This is kind of implicit

README.md
eshwar-sundar-glean: Added in f6d9c93 — the `PLUGIN_DATA_DIR` row now notes it will also hold staged arguments for write tools. Phrased as forthcoming since that storage isn't in the code yet (today the dir holds credentials, pending-auth, the remote-tools cache, the server-URL config, and the log).

README.md
eshwar-sundar-glean: It's already 5 minutes — the value is `300000` ms, not `30000`. `.mcp.json` and `.mcp.codex.json` both set `"HITL_TIMEOUT_MS": "300000"`, and the bundle default is `defaultHitlTimeoutMs = 300_000` (src/tools/run-tool.ts:13). So no code or doc change needed here — looks like a dropped zero on read.

README.md
eshwar-sundar-glean: Next PR

README.md
eshwar-sundar-glean: Done in 711f7ba. Restored the `CODEX_THREAD_ID` → `GLEAN_SESSION_ID` mapping that was dropped when the per-host launchers were unified into `plugins/glean/start.sh` — added as an `elif` after the Claude Code branch:

```sh
if [ -n "${CLAUDE_CODE_SESSION_ID:-}" ]; then
  export GLEAN_SESSION_ID="$CLAUDE_CODE_SESSION_ID"
elif [ -n "${CODEX_THREAD_ID:-}" ]; then
  export GLEAN_SESSION_ID="$CODEX_THREAD_ID"
fi
```

Also documented `CODEX_THREAD_ID` in the README's launcher-managed variables table.

README.md
```diff
@@ -23,12 +24,22 @@ Today it ships one plugin:
...
+### Codex and Cursor
 
-1. Open the plugin picker.
-2. Click **Add marketplace**, choose **GitHub**, and enter
-   `gleanwork/glean-plugins-vnext`.
-3. Once the marketplace syncs, install the **glean-vnext** plugin from it.
+The same plugin ships host-specific manifests so it can be installed from
+Codex and Cursor as well:
+
+- **Codex** reads `.agents/plugins/marketplace.json` (marketplace
```
mohit-gupta-glean: #clarify Is this detail needed? If yes, should this be part of install section?

README.md
```diff
@@ -23,12 +24,22 @@ Today it ships one plugin:
...
+  `glean-plugins-vnext`) and the per-host manifest at
+  `plugins/glean/.codex-plugin/plugin.json`, which points the MCP server at
+  `.mcp.codex.json`.
+- **Cursor** reads `.cursor-plugin/marketplace.json` (marketplace
+  `glean-plugins`) and `plugins/glean/.cursor-plugin/plugin.json`, which reuses
+  the shared `.mcp.json`.
+
+Point the host at this repo (`gleanwork/glean-plugins-vnext`) and install
```
mohit-gupta-glean: #suggest can we add more precise instructions, maybe with screenshots or point to the corresponding docs on how to add a plugin.

mohit-gupta-glean: e.g. https://cursor.com/docs/plugins#team-marketplaces

mohit-gupta-glean: It may be good to mention that the repo is the marketplace which has the plugin

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+variables below are optional and read directly by the server bundle —
+`GLEAN_MCP_SERVER_URL` overrides the URL saved by `setup`, the storage and
+session variables are normally derived by the launcher
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
+
+| Variable | Purpose | Default |
+| --- | --- | --- |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL. When unset, the URL saved by the `setup` tool is used. | (stored config from `setup`) |
+| `ENABLE_HITL` | Enables human-in-the-loop confirmation before `run_tool` executes a downstream tool. Active only when set to exactly `true`. | disabled |
```
mohit-gupta-glean: default is true now

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+| `HITL_TIMEOUT_MS` | Timeout, in milliseconds, for a human-in-the-loop confirmation prompt. Must be a positive integer. | `300000` (5 min) |
+| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size, in bytes, of each file read via `run_tool`'s `file_args`. Must be a positive integer. | `1048576` (1 MiB) |
+| `PLUGIN_DATA_DIR` | Directory for cached credentials, pending-auth state, the remote-tools cache, the saved server URL, and `glean-server.log`. Will also hold staged arguments for write tools. | `~/.glean` |
+| `SKILLS_BASE_DIR` | Directory where discovered skill files are written. | `/tmp/glean-skills-cache` |
```
mohit-gupta-glean: SKILLS_BASE_DIR is currently derived and always overridden here: https://github.com/gleanwork/glean-plugins-vnext/blob/cfdf002cfae50e485a71537da24d59be679b4085/plugins/glean/start.sh#L13

is this correct?

#Suggest drop it from documentation and let it be derived based on PLUGIN_DATA_DIR and USE_CLAUDE_PROJECT_DIR

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+Primary configuration is interactive, not environment-based: the `setup` tool
+captures your Glean Server URL and drives OAuth sign-in on first run (see
+[First run](#first-run)), persisting both under `~/.glean/`. The environment
+variables below are optional and read directly by the server bundle —
+`GLEAN_MCP_SERVER_URL` overrides the URL saved by `setup`, the storage and
+session variables are normally derived by the launcher
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
+
+| Variable | Purpose | Default |
```
mohit-gupta-glean, line 108: Let's document USE_CLAUDE_PROJECT_DIR

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+| --- | --- | --- |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL. When unset, the URL saved by the `setup` tool is used. | (stored config from `setup`) |
+| `ENABLE_HITL` | Enables human-in-the-loop confirmation before `run_tool` executes a downstream tool. Active only when set to exactly `true`. | disabled |
+| `HITL_TIMEOUT_MS` | Timeout, in milliseconds, for a human-in-the-loop confirmation prompt. Must be a positive integer. | `300000` (5 min) |
+| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size, in bytes, of each file read via `run_tool`'s `file_args`. Must be a positive integer. | `1048576` (1 MiB) |
+| `PLUGIN_DATA_DIR` | Directory for cached credentials, pending-auth state, the remote-tools cache, the saved server URL, and `glean-server.log`. Will also hold staged arguments for write tools. | `~/.glean` |
```
mohit-gupta-glean: PLUGIN_DATA_DIR is currently derived and always overwritten here: https://github.com/gleanwork/glean-plugins-vnext/blob/cfdf002cfae50e485a71537da24d59be679b4085/plugins/glean/start.sh#L35

Don't think that overrides to this are respected. If we want that, we should update start.sh. 

Otherwise, this is just an internal detail.

README.md
mohit-gupta-glean: Edit: SKILLS_BASE_DIR is not derived from PLUGIN_DATA_DIR. We should change that to first be derived from PLUGIN_DATA_DIR and then overridden if USE_CLAUDE_PROJECT_DIR = 1

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+| `SKILLS_BASE_DIR` | Directory where discovered skill files are written. | `/tmp/glean-skills-cache` |
+| `GLEAN_SESSION_ID` | Chat session id sent with backend calls. | a UUID generated once per process |
+
+Empty values and un-interpolated `${VAR}` placeholders are ignored, so a host
+that passes an unset variable through verbatim falls back to the default.
+
+The defaults above are the bundle's built-in fallbacks. The shipped MCP
+manifests (`plugins/glean/.mcp.json` and `.mcp.codex.json`) set
+`ENABLE_HITL=true` and `HITL_TIMEOUT_MS=300000` in their `env` block, so
```
mohit-gupta-glean: #suggest This is enabled by default right?

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+## Configuration
+
+Primary configuration is interactive, not environment-based: the `setup` tool
+captures your Glean Server URL and drives OAuth sign-in on first run (see
+[First run](#first-run)), persisting both under `~/.glean/`. The environment
+variables below are optional and read directly by the server bundle —
+`GLEAN_MCP_SERVER_URL` overrides the URL saved by `setup`, the storage and
+session variables are normally derived by the launcher
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
```
mohit-gupta-glean: It's not clear what the purpose of documenting these env variables is? Shouldn't we only specify variables that someone can override?

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+
+### Launcher-managed variables
+
+`start.sh` runs host-side and normalizes the following host inputs into the
+plugin variables above, keeping the server bundle host-agnostic:
+
+| Host variable | Effect |
+| --- | --- |
+| `CLAUDE_PLUGIN_DATA` | Host-managed lifecycle dir. When set, both `PLUGIN_DATA_DIR` and `SKILLS_BASE_DIR` (as `<dir>/glean-skills-cache`) are anchored under it. |
+| `USE_CLAUDE_PROJECT_DIR` | Opt-in (`=1`): routes the skills cache under the launch project's `.claude/tmp/` so the `glean_run` skill's `Read` glob can match cache files. |
```
mohit-gupta-glean: This is not really a host variable. This can be set by the users anywhere right?

README.md
mohit-gupta-glean: #clarify why do we specify some variables in mcp.json and others are simply exported from start.sh? Is this because we wanted these to be different per host.

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+session variables are normally derived by the launcher
+(`plugins/glean/start.sh`) from what the host provides, and the rest tune
+runtime behavior. None are required for normal use.
+
+| Variable | Purpose | Default |
+| --- | --- | --- |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL. When unset, the URL saved by the `setup` tool is used. | (stored config from `setup`) |
+| `ENABLE_HITL` | Enables human-in-the-loop confirmation before `run_tool` executes a downstream tool. Active only when set to exactly `true`. | disabled |
+| `HITL_TIMEOUT_MS` | Timeout, in milliseconds, for a human-in-the-loop confirmation prompt. Must be a positive integer. | `300000` (5 min) |
+| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size, in bytes, of each file read via `run_tool`'s `file_args`. Must be a positive integer. | `1048576` (1 MiB) |
```
mohit-gupta-glean: I don't see the flag being set anywhere to 1 MiB

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
+| --- | --- |
+| `CLAUDE_PLUGIN_DATA` | Host-managed lifecycle dir. When set, both `PLUGIN_DATA_DIR` and `SKILLS_BASE_DIR` (as `<dir>/glean-skills-cache`) are anchored under it. |
+| `USE_CLAUDE_PROJECT_DIR` | Opt-in (`=1`): routes the skills cache under the launch project's `.claude/tmp/` so the `glean_run` skill's `Read` glob can match cache files. |
+| `CLAUDE_CODE_SESSION_ID` | Claude Code's conversation id; copied into `GLEAN_SESSION_ID` so the session id tracks the host conversation. |
+| `CODEX_THREAD_ID` | Codex's conversation id; used as `GLEAN_SESSION_ID` when `CLAUDE_CODE_SESSION_ID` is unset. (Cursor exposes no session id, so the plugin generates one.) |
+| `HOME` | Fallback base for `PLUGIN_DATA_DIR` (`~/.glean`) and `SKILLS_BASE_DIR` when `CLAUDE_PLUGIN_DATA` is absent. |
+
 ## Troubleshooting
 
 - **Sign-in loop** — the cached OAuth provider state may be stale. Delete
```
mohit-gupta-glean: Instead of asking people to manually delete, shouldn't we just ask to run setup with reset?

README.md
```diff
@@ -85,12 +92,56 @@ For local development, point the marketplace at your local checkout instead:
...
   `~/.glean/mcp-credentials.json` and retry.
```
mohit-gupta-glean: Isn't this path derived from plugin data dir?

mohit-gupta-glean (commented):
Thanks for doing this Eshwar!

eshwar-sundar-glean:
Thanks @mohit-gupta-glean — great catches. Reworked the Configuration + Install + Troubleshooting sections in 1d69a1f. Summary of how each point landed:

**Core critique (only document what's overridable):** The env-var table now lists *only* user-overridable variables: `GLEAN_MCP_SERVER_URL`, `ENABLE_HITL`, `HITL_TIMEOUT_MS`, `GLEAN_FILE_ARG_MAX_BYTES`, and `USE_CLAUDE_PROJECT_DIR`.

**`PLUGIN_DATA_DIR` / `SKILLS_BASE_DIR` / `GLEAN_SESSION_ID` are clobbered by start.sh:** Confirmed — start.sh unconditionally exports them (L13/L28/L35), so user values don't survive. Moved them out of the table into a short note that says exactly that: "start.sh overwrites them on every launch and setting them yourself has no effect."

**`ENABLE_HITL` / `HITL_TIMEOUT_MS` default is true now:** Fixed — the table shows `ENABLE_HITL` = `true` and `HITL_TIMEOUT_MS` = `300000`, both noted as set in the shipped `.mcp.json`.

**`GLEAN_FILE_ARG_MAX_BYTES` isn't set anywhere to 1 MiB:** Right — it's the bundle default, not in `.mcp.json`. Labeled "bundle default."

**`USE_CLAUDE_PROJECT_DIR` isn't a host variable / please document it:** Agreed — promoted it into the overridable table (it's a user opt-in), and dropped the misleading "host inputs" table.

**Why some vars are in mcp.json vs start.sh:** The intro now says overridable knobs go in the MCP `env` block (the shipped manifests set the HITL ones), while the derived dir/session vars are exported by start.sh.

**Install (Codex/Cursor):** Trimmed the internal manifest detail, stated the repo *is* the marketplace, and linked Cursor's [Team marketplaces](https://cursor.com/docs/plugins#team-marketplaces) docs.

**Troubleshooting:** Sign-in-loop fix now points to `setup` with `reset=true` instead of deleting a hardcoded `~/.glean/...` path.

**One deferred item (code, not docs):** your suggestion to derive `SKILLS_BASE_DIR` from `PLUGIN_DATA_DIR` (then override on `USE_CLAUDE_PROJECT_DIR=1`) is a start.sh change. Keeping this PR docs-only; the doc reflects current behavior (anchored under `CLAUDE_PLUGIN_DATA`/`~/.claude/tmp`). Happy to do the start.sh change in a follow-up — want me to file it?

README.md
eshwar-sundar-glean: Implemented in a39d0a1 (decided to fold it into this PR rather than defer). start.sh now computes `PLUGIN_DATA_DIR` first, then derives `SKILLS_BASE_DIR="$PLUGIN_DATA_DIR/glean-skills-cache"`, with the `USE_CLAUDE_PROJECT_DIR=1` redirect applied after. The only behavior change is the no-`CLAUDE_PLUGIN_DATA` fallback: skills now live under `~/.glean/glean-skills-cache` instead of `~/.claude/tmp/...`. The skill's Read glob keys on the `glean-skills-cache` dir name, so matching is unaffected; tests green. README updated to match.

README.md
eshwar-sundar-glean: Heads up: in the Configuration restructure (Mohit's review), `PLUGIN_DATA_DIR` moved into an 'internal/derived' note and I dropped the 'staged write-tool arguments' mention for now, since that storage isn't in the code yet and the README otherwise documents current behavior. Easy to re-add the line once the write-tool arg staging actually lands under PLUGIN_DATA_DIR.

README.md
eshwar-sundar-glean: Done

README.md
eshwar-sundar-glean: start.sh can be dynamically computed, mcp server has to be statically set while shipping. thats the only difference.

plugins/glean/start.sh
swarup-padhi-glean: Nit: Modify comment to remove cowork

README.md
```diff
@@ -23,29 +24,27 @@ Today it ships one plugin:
...
 
 ## First run
 
 The first tool call triggers OAuth sign-in to Glean via the `setup` tool.
-The agent's skill walks the user through the sign-in URL and paste-back
-flow. After sign-in, OAuth credentials are cached to `~/.glean/` and reused
-across sessions — you won't be prompted again until the refresh token
+`setup` opens the Glean sign-in page in your browser and captures the
+authorization code in-context through a local loopback callback — no
+copy-paste. After sign-in, OAuth credentials are cached to `~/.glean/` and
```
swarup-padhi-glean: nit: no copy-paste seems not necessary here

README.md
```diff
@@ -85,12 +84,49 @@ For local development, point the marketplace at your local checkout instead:
...
+A few optional variables let you override behavior. Set them in the host's MCP
+server `env` block — the shipped `.mcp.json` / `.mcp.codex.json` already set the
+HITL ones — or in your shell:
+
+| Variable | Purpose | Default |
+| --- | --- | --- |
+| `GLEAN_MCP_SERVER_URL` | Overrides the Glean server URL captured by `setup`. | URL saved by `setup` |
+| `ENABLE_HITL` | Human-in-the-loop confirmation before `run_tool` runs a downstream tool. Active only when set to exactly `true`. | `true` (set in the shipped `.mcp.json`) |
+| `HITL_TIMEOUT_MS` | Timeout in milliseconds for a HITL confirmation prompt. Positive integer. | `300000` (5 min), set in `.mcp.json` |
+| `GLEAN_FILE_ARG_MAX_BYTES` | Maximum size in bytes of each file read via `run_tool`'s `file_args`. Positive integer. | `1048576` (1 MiB), bundle default |
```
swarup-padhi-glean, line 113: I am just curious how will we handle the cases when we make code changes. These values are easy to diverge from actual code.

README.md
```diff
@@ -85,12 +84,49 @@ For local development, point the marketplace at your local checkout instead:
...
+  Claude Code, `CODEX_THREAD_ID` for Codex, otherwise a generated UUID.
+
 ## Troubleshooting
 
-- **Sign-in loop** — the cached OAuth provider state may be stale. Delete
-  `~/.glean/mcp-credentials.json` and retry.
-- **`GLEAN_MCP_SERVER_URL is required`** — the plugin's `.mcp.json` wasn't
-  picked up by the host. Reinstall; if that fails, open an issue.
+- **Sign-in loop or stale auth** — call the `setup` tool with `reset=true` to
+  clear the saved configuration and credentials, then call `setup` again to
```
swarup-padhi-glean: #suggest: Shouldn't we rather frame it as prompt to reset?

README.md
```diff
@@ -85,12 +84,49 @@ For local development, point the marketplace at your local checkout instead:
...
+  re-authenticate.
+- **Tools return `[SETUP_REQUIRED]`** — the plugin isn't configured or
+  authenticated yet. Call the `setup` tool (no arguments) to advance through
+  the next missing stage: saving the Server URL, signing in, then fetching the
```
swarup-padhi-glean: #suggest: Again can we reframe to ask them re-prompt to setup?

swarup-padhi-glean (commented):
LGTM

README.md
eshwar-sundar-glean: Best we can do is add a skill

plugins/glean/start.sh
eshwar-sundar-glean: Done in bcd290e — header comment now reads `(Claude Code, Codex, or Cursor)`, Cowork dropped.

README.md
eshwar-sundar-glean: Done in bcd290e — removed the 'no copy-paste' aside.

README.md
eshwar-sundar-glean: Done in bcd290e — reframed as a user prompt while keeping the tool detail: "prompt the agent to reset and sign in again (e.g. \"reset the Glean setup\"). It calls the `setup` tool with `reset=true` …"

README.md
eshwar-sundar-glean: Done in bcd290e — same treatment: "Prompt the agent to set up Glean (e.g. \"set up Glean\"). The `setup` tool, called with no arguments, advances through the next missing stage …"

README.md
eshwar-sundar-glean: Fair concern — there's no automated sync today, so these defaults can drift (this PR was a manual reconcile against the code). The values come from a few constants (`defaultHitlTimeoutMs`, `DEFAULT_FILE_ARG_MAX_BYTES`) and the shipped `.mcp.json`. I kept the table to only user-overridable knobs to minimize surface area. If you'd like, I can add a small test that asserts the documented defaults match those constants so CI catches drift — happy to do it as a follow-up.

eshwar-sundar-glean:
@swarup-padhi-glean done and addressed your comments

swarup-padhi-glean (approved):

PR #16, opened 2026-06-20
WIP - Make connector AUTH_REQUIRED explicit in run_tool

eshwar-sundar-glean:
## Summary

Makes **downstream connector authorization** explicit in `run_tool`. When a
connector (Jira/Slack/Salesforce via the Glean gateway) needs the user to
authorize their account — *even though Glean itself is already authenticated* —
the gateway returns an `isError` result whose text is a JSON envelope with
`authUrls`. Previously the plugin passed this straight through, and the model
could under-surface it or **confuse it with the plugin's own `[SETUP_REQUIRED]`
(Glean sign-in) and wrongly call `setup`**.

`handleRunTool` now detects that result and:

1. **Always** appends a disambiguation note as an extra text block so the model
   gets the message regardless of client capabilities:
   > NOTE: This is NOT [SETUP_REQUIRED]. Glean itself is already authenticated —
   > only this downstream tool/connector needs authorization. Do NOT call
   > `setup`. Show the link(s) above, have the user authorize, then retry this
   > tool.
2. **If the client supports elicitation**, fires a single best-effort
   form/confirm dialog surfacing the connector-auth requirement (accept /
   decline / timeout are ignored — the suffixed result is returned regardless).
3. Is **always-on** — not gated by `ENABLE_HITL` (this is surfacing, not an
   approval gate).

Normal results and ordinary (non-`authUrls`) errors pass through unchanged.

## Detection

`result.isError === true` AND the first text content parses as JSON with a
non-empty `authUrls: string[]`. Reuses the existing elicitation gate
(`getClientCapabilities()?.elicitation`) and `hitlTimeoutMs()` from the HITL
path.

**Follow-up (noted, not in this PR):** the gateway also sets an `AUTH_REQUIRED`
marker on the result `_meta`; keying on that would be more robust (and would
catch the no-`authUrls` fallback message). The exact `_meta` field name needs
confirming from a live response.

## Changes

- `src/tools/run-tool.ts` — `parseConnectorAuth`, `withConnectorAuthSuffix`,
  `connectorAuthPrompt`, `CONNECTOR_AUTH_SUFFIX`; `handleRunTool` tail now
  captures the result, surfaces connector auth, and returns the suffixed result.
- `tests/run-tool.test.ts` — 5 new cases (elicit + suffix; no-elicitation client;
  declined/errored dialog still suffixed; normal result unchanged; non-JSON error
  unchanged).
- Plugin manifests bumped `0.2.26 → 0.2.27` (aligned); dist rebuilt.

## Testing

- `npm test` — 141/141 passing.
- `npm run build` — dist in sync (esbuild 0.28.1).
- `bash scripts/check-version-bump.sh origin/main` — `0.2.26 → 0.2.27 (all manifests aligned)`.

### DIFF
--- a/src/tools/run-tool.ts
+++ b/src/tools/run-tool.ts
@@ -217,7 +217,7 @@ function connectorAuthPrompt(toolName: string): string {
   return (
     `"${toolName}" needs you to authorize its connector before it can run. ` +
     `The assistant will show the sign-in link(s) — open them, then ask it to ` +
-    `retry. (This is separate from Glean setup; Glean is already connected.)`
+    `retry.`
   );
 }
 
--- a/src/tools/run-tool.ts
+++ b/src/tools/run-tool.ts
@@ -170,6 +170,66 @@ function buildApprovalMessage(
   return [`Action: ${toolName}`, "Arguments:", formatArguments(args)].join("\n");
 }
 
+// Appended to a connector AUTH_REQUIRED result so the model surfaces the
+// connector sign-in and does not confuse it with the plugin's own
+// [SETUP_REQUIRED] (Glean sign-in). Plain text — clients don't reliably render
+// markdown here.
+const CONNECTOR_AUTH_SUFFIX =
+  "NOTE: This is NOT [SETUP_REQUIRED]. Glean itself is already authenticated " +
+  "— only this downstream tool/connector needs authorization. Do NOT call " +
+  "`setup`. Show the link(s) above, have the user authorize, then retry this " +
+  "tool.";
+
+// Detect the gateway's connector AUTH_REQUIRED envelope: an error result whose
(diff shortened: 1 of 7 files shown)


src/tools/run-tool.ts
```diff
@@ -170,6 +170,66 @@ function buildApprovalMessage(
...
+    );
+    if (authUrls.length > 0) return { authUrls };
+  }
+  return null;
+}
+
+// Plain-text heads-up shown to the user when a connector needs authorization.
+// No clickable URLs (clients don't render markdown in elicitation; the spec
+// discourages URLs in form fields) — the links come from the result text.
+function connectorAuthPrompt(toolName: string): string {
```
eshwar-sundar-glean: ```(This is separate from Glean setup; Glean is already connected.)```
This part is not required

PR #17, opened 2026-06-20
run_tool - Clarify Glean is already authenticated on connector AUTH_REQUIRED

eshwar-sundar-glean:
## Summary

Minimal, **prompt-only** handling of downstream connector authorization in
`run_tool`. When the Glean gateway returns a connector AUTH_REQUIRED result
(`isError` + a JSON envelope with `authUrls`), append a short clarification
text block:

> Note: Glean itself is already authenticated. This is a downstream
> connector/tool authorization request — NOT [SETUP_REQUIRED]. Do not call the
> `setup` tool. Have the user authorize using the link above, then retry this
> tool.

This stops the model from confusing connector auth with the plugin's own
`[SETUP_REQUIRED]` (Glean sign-in) and wrongly calling `setup`.

### DIFF
--- a/src/tools/run-tool.ts
+++ b/src/tools/run-tool.ts
@@ -170,14 +170,13 @@ function buildApprovalMessage(
   return [`Action: ${toolName}`, "Arguments:", formatArguments(args)].join("\n");
 }
 
-// Appended to a connector AUTH_REQUIRED result so the model understands the
-// downstream connector (not Glean) needs authorization, and does not confuse it
-// with the plugin's own [SETUP_REQUIRED] Glean sign-in.
+// Instruction to the ASSISTANT (not user-facing) appended to a connector
+// AUTH_REQUIRED result: keep the user-facing ask minimal and don't leak auth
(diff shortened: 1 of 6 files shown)


src/tools/run-tool.ts
```diff
@@ -170,6 +170,44 @@ function buildApprovalMessage(
   return [`Action: ${toolName}`, "Arguments:", formatArguments(args)].join("\n");
 }
 
+// Instruction to the ASSISTANT (not user-facing) appended to a connector
+// AUTH_REQUIRED result: keep the user-facing ask minimal and don't leak auth
+// internals, while still preventing a wrong `setup` call.
+const CONNECTOR_AUTH_SUFFIX =
+  "Assistant: ask the user only to authorize using the link above. Do not " +
```
eshwar-sundar-glean, line 177: Without giving a prompt like this, the LLM keeps saying stuff like
`I won't call setup`,
`We already authenticated to Glean, now connectors`
etc.

PR #18, opened 2026-06-20
[Plugins] Surface file_args content in the HITL approval prompt

swarup-padhi-glean:
## Problem

`run_tool` accepts `file_args` (arg name → absolute file path); the plugin reads each file and merges its content into `arguments` before calling the action. That resolution happened **after** the HITL approval, so file-sourced values never appeared in the approval prompt — a user could approve an action (e.g. "send a message") without seeing input that comes from a file (the message body sourced from `file_args`).

## Fix

Resolve `file_args` **before** building the approval prompt and build the prompt from the merged result. File-sourced values now appear in the displayed arguments alongside inline ones; large file content is truncated inline and written in full to the spill file, same as any large argument. An unreadable `file_args` path now also fails **before** prompting (instead of after the user approves).

## Stacked on the HITL display PR

Based on the HITL display PR rather than `main`: surfacing (potentially large) file content stays viewport-safe only because of that PR's truncate-inline + spill-to-file behavior — on `main` alone a large file body would overflow the prompt. When the base PR merges, this retargets to `main` automatically.

## Tests

- file_args content appears in the approval message; the action executes on accept.
- an unreadable file_args path fails before prompting (no elicitation, not executed).
- Build + typecheck + tests all green.

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.27",
+  "version": "0.2.28",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.27",
+  "version": "0.2.28",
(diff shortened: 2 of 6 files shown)


swarup-padhi-glean:
Before:
Screenshot 2026-06-21 at 12 23 46 PM
Screenshot 2026-06-21 at 12 25 31 PM


After:
Screenshot 2026-06-21 at 11 51 00 AM
Screenshot 2026-06-21 at 12 01 24 PM

eshwar-sundar-glean (approved):

PR #23, opened 2026-06-22
[Plugins] Auto-approve run_tool's native prompt under HITL (Claude Code PreToolUse hook)

swarup-padhi-glean:
## What & why
When HITL is enabled, the Glean plugin's own elicitation prompt is the approval gate for `run_tool`. Claude Code still shows its **own** native "Allow this tool?" prompt on top of it — a redundant double-prompt. This adds a Claude Code **PreToolUse hook** that auto-approves `run_tool`'s native prompt when HITL is active, leaving the HITL elicitation as the single gate.

## How it's gated (safety)
The hook emits `permissionDecision: "allow"` only when **both**:
- the tool is **Glean's `run_tool`** (the tool name carries `glean` and resolves to `run_tool`), and
- **`ENABLE_HITL=true`** (read from the plugin's `.mcp.json`, since the hook process doesn't inherit the MCP server env).

`run_tool` is read-only *only* while HITL gates it. This hook runs only under Claude Code, which always advertises the `elicitation` capability, so `ENABLE_HITL=true` ⇒ `run_tool`'s HITL prompt is active — it can never auto-approve an ungated write. When `ENABLE_HITL` isn't `true`, the hook does nothing and the normal permission flow runs. A stricter `deny`/`ask` from any other PreToolUse hook still wins.

## Scope
- **Claude Code only.** Cursor/Codex also have PreToolUse hooks, but they don't surface the redundant native prompt, so there's nothing to suppress there.
- **`run_tool` only** for now. Extending to other read-only tools (`find_skills`, `setup`) and other surfaces is a follow-up.

## Also
- Fixes a duplicate-hooks load error: removed the redundant `hooks` field from the Claude manifest (the host auto-loads `hooks/hooks.json`; declaring it again double-loaded and failed).

## Tests
Unit tests for the hook: allows Glean `run_tool` when HITL on; never allows when HITL off (safety); ignores a non-Glean `run_tool`; ignores Glean tools other than `run_tool`.


plugins/glean/hooks/auto-approve-run-tool.mjs
```diff
@@ -0,0 +1,60 @@
...
+let input = {};
+try {
+  input = JSON.parse(readStdin());
+} catch {
+  // Malformed/empty input: do nothing, let the normal permission flow run.
+}
+
+const toolName = String(input.tool_name ?? "");
+// Only this plugin's run_tool meta-tool (exposed as mcp__<server>__run_tool).
+if (!toolName.endsWith("run_tool")) process.exit(0);
```
eshwar-sundar-glean: Check for both glean and run_tool?

plugins/glean/hooks/auto-approve-run-tool.mjs
```diff
@@ -0,0 +1,60 @@
...
+try {
+  const root = process.env.CLAUDE_PLUGIN_ROOT ?? ".";
+  const cfg = JSON.parse(fs.readFileSync(path.join(root, ".mcp.json"), "utf-8"));
+  env = cfg?.mcpServers?.glean?.env ?? {};
+} catch {
+  // No readable config: do nothing.
+}
+
+const hitlOn = env.ENABLE_HITL === "true";
+const autoApprove = env.HITL_AUTO_APPROVE === "true";
```
eshwar-sundar-glean: Dont think we need a new flag here, just use ENABLE_HITL

plugins/glean/.claude-plugin/plugin.json
```diff
@@ -1,8 +1,9 @@
 {
```
eshwar-sundar-glean, line 1: Can you check what does it mean to support it in Codex and Cursor as well?

Cursor - https://cursor.com/docs/hooks#pretooluse
Codex - https://developers.openai.com/codex/hooks#pretooluse

eshwar-sundar-glean (approved):
You can merge now and take up other surfaces in next PR if needed, it will be good to maintain consistency across surfaces where applicable

plugins/glean/.claude-plugin/plugin.json
swarup-padhi-glean: Hey @eshwar-sundar-glean we don't have this problem in cursor/codex. We don't need to solve there.

PR #24, opened 2026-06-22
find_skills: guide re-discovery / in-scope tools when a skill has no tools

eshwar-sundar-glean:
When find_skills returns a skill that resolves to **no tools** (e.g. tool-less UGC skills or servers:[] static skills), the model otherwise gets just the skill's playbook with no signal about what to do next. This adds guidance to the `find_skills` tool description: if a returned skill lists no tools and its playbook can't complete the task, first check tools already in scope — from other skills in this response, earlier find_skills calls, or tools you can already call directly — and only re-discover with reworded or additional queries if none fit.


anirudh-bijay-glean (approved):

mohit-gupta-glean (commented):
LGTM

PR #27, opened 2026-06-24
Auto-dismiss HITL approval popup on elicitation timeout

pragati-agrawal-glean:
When a user doesn't respond to a HITL approval prompt within HITL_TIMEOUT_MS, the popup got stuck: the user had to press Escape and retry. Root cause is a client-side MCP SDK bug. The first server->client request in a session gets JSON-RPC id 0. The SDK's `_oncancel` bails on a falsy requestId (`if (!notification.params.requestId) return`), and `!0` is true — so the `notifications/cancelled` we send on timeout for request id 0 is silently dropped by the client, and the accept/decline popup never auto-dismisses (it lingers with dead buttons). Prompts 2+ get id>=1 and dismiss fine, which is why it looked flaky ("sometimes it disappears").

Fix: burn id 0 with a throwaway ping before the first elicitation, so every approval prompt uses id>=1 and its timeout cancellation is honored. The id is consumed synchronously inside request(), so it's fire-and-forget and tolerant of ping failure. The real fix belongs in the SDK's `_oncancel`; this guards us until clients ship it.
(Reference : https://github.com/modelcontextprotocol/typescript-sdk/issues/2283) 

- src/tools/run-tool.ts: primeElicitationCancellation(), once per server
- tests/run-tool.test.ts: cover the burn (pings once; not when no approval)
- plugins/glean/dist/index.js: rebuilt
- bump all 3 plugin manifests 0.2.33 -> 0.2.34

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.34",
+  "version": "0.2.35",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.34",
+  "version": "0.2.35",
   "description": "Glean Codex plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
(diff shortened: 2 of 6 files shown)


swarup-padhi-glean (commented):
Can you also post screenshots of its working in the slack channel?

pragati-agrawal-glean:
> Can you also post screenshots of its working in the slack channel?

It works the same way as the third video added in the slack channel on the thread 
https://askscio.slack.com/files/U0B8TT25351/F0BBW13BESK/automatic.mov

swarup-padhi-glean (approved):
LGTM, good catch. cc: @mohit-gupta-glean

eshwar-sundar-glean (approved):

mohit-gupta-glean (commented):
LGTM

PR #33, opened 2026-06-29
Fix version bump pre-commit hook

anirudh-bijay-glean:
## What
Update the pre-commit hook for automatic version bump (`.githooks/pre-commit.d/02-auto-bump-version.sh`) to update `plugins/glean/.codex-plugin/plugin.json` and `plugins/glean/.cursor-plugin/plugin.json` as well.

## Why
The pre-commit hook was previously updating only `plugins/glean/.codex-plugin/plugin.json`.

## Details
The script keeps the versions across the plugin manifests in lockstep. The larger number of possible cases here makes determining the target version complex. The target version is computed as follows:

Let $V_\rm{main}$ be the set of all valid versions (across all manifests) in the `main` branch and $V_\rm{current}$ be the set of all valid version strings taking into consideration the staged changes. Then,
- if both $V_\rm{main}$ and $V_\rm{current}$ are empty, the script errors; else
- if $V_\rm{main} = \emptyset$, $v = \rm{max} ~ V_\rm{current}$; else
- if $V_\rm{current} = \emptyset$, $v = \rm{max}$ { $u + 0.0.1 \mid u \in V_\rm{main}$ }; else
- $v = \rm{max}$ { $\rm{max} ~ V_\rm{current}, \rm{max}$ { $u + 0.0.1 \mid u \in V_\rm{main}$ } }
where $v$ is the target version, $\rm{max}$ is defined relative to semver comparison, and $u + 0.0.1$ increments the patch version of $u$ by 1.

Versions are assumed to be of the form $x.y.z$ (no pre-release tags/build metadata).

~Missing plugin manifests do not fail the hook; however, malformed JSON in the plugin manifests does.~
Missing plugin manifests and malformed JSON in the plugin manifests fail the hook.

## Code changes
- Update `02-auto-bump-version.sh`.
- Update the shebang of shell scripts (`01-rebuild-dist.sh`, `02-auto-bump-version.sh`, `check-version-bump.sh`, `pack-plugin.sh`) to use the environment Bash instead of the system Bash. (`start.sh` has been omitted.)

### DIFF
--- a/.githooks/pre-commit.d/02-auto-bump-version.sh
+++ b/.githooks/pre-commit.d/02-auto-bump-version.sh
@@ -1,44 +1,108 @@
-#!/bin/bash
-# Auto-bump patch version if plugin files are staged and version wasn't bumped.
-# Uses the same path regex as scripts/check-version-bump.sh (CI check).
+#!/usr/bin/env bash
+# Auto-bump patch version across all plugin manifests in lockstep when plugin files
+# are staged. Uses the same path regex as scripts/check-version-bump.sh (the CI gate).
+#
+# The new version, v_new, is determined as follows:
+#
+#     v_main = max {version of each manifest on main}
+#     v_current = max {version of each manifest in the working tree}
+#     v_new = max {v_current, v_main + 0.0.1}
+#
+# where max is defined with respect to semver comparison and v_main + 0.0.1 is the patch
+# increment of v_main. Versions are plain x.y.z (no pre-release or build metadata). A
+# manifest whose version is not a plain x.y.z string is dropped from the max computation
+# but still rewritten to v_new; a manifest that is not valid JSON aborts the commit with
+# a message.
 set -e
 cd "$(git rev-parse --show-toplevel)"
 
-PLUGIN_JSON="plugins/glean/.claude-plugin/plugin.json"
-PLUGIN_PATHS="^(src/|plugins/glean/(dist/|skills/|start\.sh|\.mcp\.json|package\.json|\.claude-plugin/plugin\.json))"
+BASE_REF="origin/main"
+
+# Per-host plugin manifests — kept in lockstep on a single version.
+MANIFESTS=(
+  "plugins/glean/.claude-plugin/plugin.json"
+  "plugins/glean/.codex-plugin/plugin.json"
+  "plugins/glean/.cursor-plugin/plugin.json"
+)
+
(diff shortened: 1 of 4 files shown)


.githooks/pre-commit.d/02-auto-bump-version.sh
```diff
@@ -1,44 +1,108 @@
...
+    const v = pkg.version;
+    if (typeof v === "string" && /^[0-9]+\.[0-9]+\.[0-9]+$/.test(v)) process.stdout.write(v);
+  ' 2>/dev/null)
+  [ -n "$v" ] && base_versions+=("$v")
+done
+
+# Versions in the working tree. A missing manifest is skipped (e.g. a platform being
+# removed). Invalid JSON aborts the commit with a clear message. A valid-JSON manifest
+# whose version is not a plain x.y.z string stays in the write set (so it gets rewritten
+# to the target) but is dropped from the max so it cannot poison the result.
```
swarup-padhi-glean: can we error out missing manifest? The expectation in my opinion is that such decision will be complemented by appropriate changes including this bump script. I would suggest let's error out when either main or working tree has it missing.

swarup-padhi-glean (approved):

.githooks/pre-commit.d/02-auto-bump-version.sh
anirudh-bijay-glean: Good point! I'd say let's error on missing manifests in the incoming changes but not on `main`. We're using a common set of manifests to check for both incoming and `main`, so if I add a manifest in the incoming changes the hook would look for the manifest in `main` as well. It should then gracefully ignore it. If I'm deleting a manifest – yes, we should error on it.

swarup-padhi-glean: Let's error on both addition and deletion as discussed offline.

anirudh-bijay-glean:
Addressed @swarup-padhi-glean's comments. The hook now errors on missing manifests in both `main` and the branch after staged changes. A couple of points to note:

**Deleting a manifest:** When deleting a manifest, the hook's `MANIFESTS` list must be updated before or along with deletion of the manifest file.

**Adding a manifest:** The manifest must be added and the change committed and merged to `main` first; only after this should the hook be updated. You may alternatively bypass the hook if the change is atomic, _i.e._, the plugin `dist` does not change (so that the hook for keeping the `dist` up-to-date need not run).

swarup-padhi-glean (approved):

PR #34, opened 2026-06-29
Parse object/array file_args as JSON via the downstream tool schema

swarup-padhi-glean:
## Problem

`file_args` always injected file contents as a **UTF-8 string**. When the target parameter is typed `object`/`array` in the downstream tool's schema (e.g. `save_agent.spec`), that string fails remote validation with `Expected object, given string`. So large structured arguments that can't be inlined — the ~140 KB agent spec reported in `#feedback-tesseract` — had **no working path**.

## Fix

Make `file_args` **schema-aware**, using the tool JSON the model already reads. For each `file_args` entry, look up the parameter's declared type at `inputSchema.properties[<arg>].type` (single string, or array form like `["object","null"]`):

- `object` / `array` → `JSON.parse` the file and inject structured data.
- anything else (the common case: doc bodies, message text) → inject the verbatim UTF-8 string — **unchanged** behavior.

Invalid JSON for an `object`/`array`-only param now fails fast with a clear `FileArgsError` instead of the opaque downstream `Expected object, given string`; a `string | object` union falls back to the raw string.

## Relationship to HITL: none (this is the part to clarify)

The parsing is **completely independent of `ENABLE_HITL`** — it behaves identically whether HITL is on or off. `findToolJson` (which loads the `inputSchema`) is hoisted to the top of `handleRunTool`, so `file_args` resolution runs on **every** `run_tool` call, *before* the HITL approval gate is even evaluated. There is no code path where parsing depends on HITL state.

For context on the original report, the two symptoms surfaced on the same `save_agent` call but are **separate** bugs:
1. the Cursor elicitation prompt not rendering / `-32001` timeout — a **HITL** issue (addressed separately). `ENABLE_HITL=false` was the reporter's workaround for *that*.
2. `Expected object, given string` — **this** bug, fixed here.

With this change, neither `ENABLE_HITL=false` nor hand-parsing `.json` file_args is needed — and turning HITL back on does not change file_args behavior.

**Known limitation (intentional):** params declared via `anyOf`/`oneOf` with no top-level `type` fall back to raw string. `save_agent.spec` uses `type: "object"`, so the reported case is covered.

## Testing

- `npm run typecheck`, `npm test` (168 passing), `npm run build` — all green. 8 new tests: object/array parsing, array-form `type`, string-param stays raw, no-schema backward compat, invalid-JSON error, `string | object` union fallback, and an end-to-end `handleRunTool` test forwarding a parsed object.
- Reproduced live on the released build (`read_document` array param via `file_args` → downstream rejected the string), then verified the fix on the real compiled bundle against a production tool-JSON fixture. The fixture test exercises the no-HITL path (no elicitation capability), and the hoist guarantees the same parse runs with HITL enabled.

All 3 host manifests aligned at 0.2.38.

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.35",
+  "version": "0.2.36",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.35",
+  "version": "0.2.36",
   "description": "Glean Codex plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.cursor-plugin/plugin.json
+++ b/plugins/glean/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "glean-vnext",
   "displayName": "Glean vNext",
-  "version": "0.2.35",
+  "version": "0.2.36",
   "description": "Search and act across your company's apps — Jira, Slack, Salesforce, Google Workspace, and more — without leaving Cursor.",
   "author": {
     "name": "Glean"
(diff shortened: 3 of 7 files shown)


eshwar-sundar-glean (commented):
@swarup-padhi-glean can you clarify how this works with hitl disabled?

eshwar-sundar-glean (approved):
LGTM , maybe update PR description, it feels as if this stuff works with HITL disabled

garvit-scio:
hmm, i had changed the agent tools to accept utf-8 strings for this reason

swarup-padhi-glean:
Tested @eshwar-sundar-glean 
Screenshot 2026-06-30 at 9 44 51 PM

PR #35, opened 2026-07-01
Claude Code - Skip run_tool HITL elicitation in bypassPermissions mode

eshwar-sundar-glean:
## What

When Claude Code is launched with `--dangerously-skip-permissions` (`permission_mode: "bypassPermissions"`), the user has opted out of every approval prompt for the session. The plugin's own HITL elicitation popup is then a redundant second gate — and, worse, when that elicitation times out it currently blocks the `run_tool` call entirely.

This change makes the plugin **skip its elicitation gate and execute directly** when — and only when — the session is in `bypassPermissions` mode. Scoped deliberately to `run_tool` and to `bypassPermissions` alone.

## Why it needs a hook → server handoff

The elicitation popup is fired by the **MCP server** (`handleRunTool` → `mcpServer.elicitInput`), but the live permission mode is only visible to the **PreToolUse hook** (via its stdin `permission_mode` field). There is no env var or MCP `initialize` field that carries the permission mode to the server, and the mode can change mid-session — so the hook is the only channel.

## How

- **`plugins/glean/hooks/auto-approve-run-tool.mjs`** — on every glean `run_tool` call (while `ENABLE_HITL` is on), writes `{ permission_mode, ts }` to `<CLAUDE_PLUGIN_DATA|~/.glean>/glean-hitl-mode/<sessionId>.json`. Best-effort; a write failure never affects the existing auto-approve decision.
- **`src/tools/run-tool.ts`** — before eliciting, reads that per-session marker; if the mode is `bypassPermissions`, skips the popup and calls the tool directly.

The base dir + session-id sanitization are kept in lockstep across the hook, `run-tool.ts`, and `start.sh`'s `PLUGIN_DATA_DIR` export.

## Safety

- **Ordering:** PreToolUse strictly precedes the tool call, so the marker is always fresh for the call being gated.
- **Resume:** a session first launched with `--dangerously-skip-permissions` and later resumed **without** it (same session id) has its leftover `bypassPermissions` marker **overwritten** with the resumed mode by the hook on the resumed session's first `run_tool` call — before the server reads it — so the gate re-engages. No stale-marker leak across resume.
- **Fail-safe:** a missing / malformed marker, or a session-id mismatch, resolves to "unknown" → the gate is **kept** (fails toward prompting, never toward a silent bypass).
- **Session-keyed:** a bypass in one session can't suppress prompts in another concurrent session.

## Testing

- `npm run typecheck` — clean
- `npm test` — 179 pass, incl. server cases (bypass skips elicit; `default` / absent / other-session all still elicit) and hook cases (marker written / session-keyed / sanitized; overwrites a leftover bypass marker on resume; nothing written when HITL off, non-glean, or mode absent).
- `npm run build` — `dist/index.js` rebuilt.

## Verification still worth doing end-to-end

Two assumptions the implementation is safe against but that confirm the feature actually fires:
1. PreToolUse hooks run under `--dangerously-skip-permissions` (strongly implied by the docs; `bypassPermissions` is a documented hook `permission_mode` value).
2. The hook's `session_id` equals the `CLAUDE_CODE_SESSION_ID` that `start.sh` feeds the server as `GLEAN_SESSION_ID`.

Both are confirmed by running once with the flag and checking that a marker lands under `~/.glean/glean-hitl-mode/`.

### DIFF
--- a/plugins/glean/hooks/auto-approve-run-tool.mjs
+++ b/plugins/glean/hooks/auto-approve-run-tool.mjs
@@ -12,6 +12,7 @@
 // write. When ENABLE_HITL is not "true" the hook does nothing and the normal
 // permission flow runs.
 import fs from "node:fs";
+import os from "node:os";
 import path from "node:path";
 
 function readStdin() {
@@ -49,6 +50,34 @@ try {
 }
 
 if (env.ENABLE_HITL === "true") {
+  // Record Claude Code's live permission mode so the MCP server can skip its
+  // own elicitation gate when the user launched with
+  // --dangerously-skip-permissions (permission_mode "bypassPermissions").
+  // Written on every run_tool call and keyed by session id, so it is always
+  // fresh for the call that immediately follows and never leaks across
+  // sessions. The base dir and session-id sanitization MUST match
+  // run-tool.ts (permissionModeMarkerPath) and start.sh's PLUGIN_DATA_DIR:
+  // CLAUDE_PLUGIN_DATA when set, else ~/.glean. Best-effort — marker I/O must
+  // never break the approval decision below.
+  try {
+    const permissionMode = String(input.permission_mode ?? "");
+    const sessionId = String(input.session_id ?? "")
+      .replace(/[^a-zA-Z0-9_-]/g, "-")
+      .slice(0, 64);
+    if (permissionMode && sessionId) {
+      const base =
+        process.env.CLAUDE_PLUGIN_DATA || path.join(os.homedir(), ".glean");
+      const dir = path.join(base, "glean-hitl-mode");
+      fs.mkdirSync(dir, { recursive: true });
+      fs.writeFileSync(
+        path.join(dir, `${sessionId}.json`),
+        JSON.stringify({ permission_mode: permissionMode, ts: Date.now() }),
+      );
+    }
+  } catch {
+    // Ignore: a failed marker write just means the server keeps prompting.
+  }
+
   process.stdout.write(
     JSON.stringify({
       hookSpecificOutput: {
--- a/src/tools/run-tool.ts
+++ b/src/tools/run-tool.ts
@@ -244,14 +244,13 @@ function primeElicitationCancellation(mcpServer: Server): void {
 
 // Path to the per-session permission-mode marker the PreToolUse hook writes
 // immediately before each run_tool call (see hooks/auto-approve-run-tool.mjs).
-// The base dir and session-id sanitization MUST stay in lockstep with the hook
-// and with start.sh's PLUGIN_DATA_DIR export, or the server reads a path the
-// hook never wrote and silently falls back to prompting.
+// This resolution MUST match the hook's exactly. The hook cannot see the
+// server-only PLUGIN_DATA_DIR that start.sh derives, so both sides key off
+// CLAUDE_PLUGIN_DATA (falling back to ~/.glean) — the one anchor available to
+// both processes. Under start.sh, PLUGIN_DATA_DIR resolves to this same path.
 function permissionModeMarkerPath(): string {
   const base =
-    process.env.PLUGIN_DATA_DIR ||
-    process.env.CLAUDE_PLUGIN_DATA ||
-    path.join(os.homedir(), ".glean");
+    process.env.CLAUDE_PLUGIN_DATA || path.join(os.homedir(), ".glean");
   const sessionId = resolveSessionId()
     .replace(/[^a-zA-Z0-9_-]/g, "-")
     .slice(0, 64);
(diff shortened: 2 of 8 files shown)


plugins/glean/hooks/auto-approve-run-tool.mjs
```diff
@@ -49,6 +50,34 @@ try {
...
+      .replace(/[^a-zA-Z0-9_-]/g, "-")
+      .slice(0, 64);
+    if (permissionMode && sessionId) {
+      const base =
+        process.env.CLAUDE_PLUGIN_DATA || path.join(os.homedir(), ".glean");
+      const dir = path.join(base, "glean-hitl-mode");
+      fs.mkdirSync(dir, { recursive: true });
+      fs.writeFileSync(
+        path.join(dir, `${sessionId}.json`),
+        JSON.stringify({ permission_mode: permissionMode, ts: Date.now() }),
```
swarup-padhi-glean, line 74: curious why we store a timestamp here as it doesn't seem to be used? One small issue I thought was that we don't clean these files up, so was it intended for helping that based on some heuristics on the timestamp?

swarup-padhi-glean (approved):
Curious why we store a timestamp field in the marker

plugins/glean/hooks/auto-approve-run-tool.mjs
eshwar-sundar-glean: I wanted these as debug logs based on what is in the file. Its a one liner and Claude code itself does not clean up session files. So not sure if its like really necessary to clean up.

src/tools/run-tool.ts
```diff
@@ -240,6 +242,38 @@ function primeElicitationCancellation(mcpServer: Server): void {
...
+// Path to the per-session permission-mode marker the PreToolUse hook writes
+// immediately before each run_tool call (see hooks/auto-approve-run-tool.mjs).
+// The base dir and session-id sanitization MUST stay in lockstep with the hook
+// and with start.sh's PLUGIN_DATA_DIR export, or the server reads a path the
+// hook never wrote and silently falls back to prompting.
+function permissionModeMarkerPath(): string {
+  const base =
+    process.env.PLUGIN_DATA_DIR ||
+    process.env.CLAUDE_PLUGIN_DATA ||
+    os.tmpdir();
```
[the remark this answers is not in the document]
eshwar-sundar-glean: Fixed in cf28641 — `permissionModeMarkerPath()` now falls back to `path.join(os.homedir(), ".glean")` instead of `os.tmpdir()`, matching the hook's write path (and start.sh's `PLUGIN_DATA_DIR` export) so the server's read path and the hook's write path stay in lockstep when neither `PLUGIN_DATA_DIR` nor `CLAUDE_PLUGIN_DATA` is set. Thanks!

src/tools/run-tool.ts
```diff
@@ -240,6 +242,46 @@ function primeElicitationCancellation(mcpServer: Server): void {
...
 }
 
+// Path to the per-session permission-mode marker the PreToolUse hook writes
+// immediately before each run_tool call (see hooks/auto-approve-run-tool.mjs).
+// The base dir and session-id sanitization MUST stay in lockstep with the hook
+// and with start.sh's PLUGIN_DATA_DIR export, or the server reads a path the
+// hook never wrote and silently falls back to prompting.
+function permissionModeMarkerPath(): string {
+  const base =
+    process.env.PLUGIN_DATA_DIR ||
```
mohit-gupta-glean: #blocking we are not checking the plugin_data_dir in the write path. Remove it from here too. Assuming that it is not available in the hook.

mohit-gupta-glean (approved):

src/tools/run-tool.ts
eshwar-sundar-glean: Done in 42666f1. Dropped `PLUGIN_DATA_DIR` from the read path so `permissionModeMarkerPath()` now uses `process.env.CLAUDE_PLUGIN_DATA || path.join(os.homedir(), ".glean")` — the exact same expression as the hook's write path. You're right that `PLUGIN_DATA_DIR` isn't available to the hook; `start.sh` only derives it for the server from `CLAUDE_PLUGIN_DATA`, so keying both sides off `CLAUDE_PLUGIN_DATA` keeps them symmetric with no behavior change under start.sh.

src/tools/run-tool.ts
```diff
@@ -244,14 +244,13 @@ function primeElicitationCancellation(mcpServer: Server): void {
...
+// This resolution MUST match the hook's exactly. The hook cannot see the
+// server-only PLUGIN_DATA_DIR that start.sh derives, so both sides key off
+// CLAUDE_PLUGIN_DATA (falling back to ~/.glean) — the one anchor available to
+// both processes. Under start.sh, PLUGIN_DATA_DIR resolves to this same path.
 function permissionModeMarkerPath(): string {
   const base =
-    process.env.PLUGIN_DATA_DIR ||
-    process.env.CLAUDE_PLUGIN_DATA ||
-    path.join(os.homedir(), ".glean");
+    process.env.CLAUDE_PLUGIN_DATA || path.join(os.homedir(), ".glean");
```
eshwar-sundar-glean, line 253: For now this is clean because this is only applicable to claude and cursor is not regressed because of this

PR #37, opened 2026-07-14
[Plugin] Fix Windows OAuth sign-in: authorize URL truncated by cmd.exe (invalid_client)

steve-calvert-glean:
## Summary

On Windows, the plugin's OAuth sign-in fails immediately with `invalid_client` / "The requested OAuth 2.0 Client does not exist", even though Dynamic Client Registration succeeds. Root cause is `openBrowser()` launching the authorize URL through `cmd /c start`.

## Root cause

The MCP SDK builds the authorize URL with `response_type` first, then `client_id` and the rest, joined by `&`:

```
/oauth/authorize?response_type=code&client_id=...&redirect_uri=...&code_challenge=...&state=...
```

`openBrowser()` passed this to `spawn("cmd", ["/c", "start", "", url])`. Node only quotes an argv element on Windows if it contains a space/tab (or is empty); the URL has neither, so it reaches `cmd.exe` unquoted. `cmd.exe` treats `&` as a command separator, so it launches only:

```
start "" .../oauth/authorize?response_type=code
```

and everything after the first `&` (client_id, redirect_uri, PKCE, state) is dropped. The browser opens a URL with no `client_id`, and the server correctly rejects it as `invalid_client`.

Observed on an affected tenant: the load-balancer log shows the browser requesting exactly `GET /oauth/authorize?response_type=code` (truncated at the first `&`), immediately after a successful `POST /oauth/register` (201). macOS/Linux are unaffected because they use `execFile("open"/"xdg-open", [url])`, which passes the URL as a single argv element with no shell.

## Fix
Eshwar - Prefer CMD for now as that was the only approach I was able to test

Launch the URL via `explorer.exe` instead of `cmd /c start`. `explorer.exe` opens the default browser and receives the URL as a single argv element with no shell, so `&` and percent-encoded characters survive intact.

Chose `explorer.exe` over `rundll32 url.dll,FileProtocolHandler` (also cmd-free) because `rundll32` is a common LOLBin frequently blocked by enterprise EDR/AppLocker, and the affected users are on managed corporate Windows. Kept `cmd` out of the path entirely rather than escaping `&`/`%`, which is fragile (percent-encoded values such as `redirect_uri=http%3A%2F%2F...` can also trip `cmd` variable expansion).

## Testing

- New `tests/open-browser.test.ts`: asserts the Windows launcher is not `cmd`, the full URL (including everything after the first `&`) is passed as one argument, and macOS/Linux behavior is unchanged.
- `npx vitest run` — 26 passed (3 new + 23 existing auth-provider tests).
- `npx tsc --noEmit` — clean.

## Please verify before merge

I validated via unit test + typecheck on macOS. Needs a manual smoke test on a real Windows machine: run the plugin sign-in and confirm the browser opens the full `?response_type=code&client_id=...` URL and OAuth completes.

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.39",
+  "version": "0.2.40",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.39",
+  "version": "0.2.40",
   "description": "Glean Codex plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.cursor-plugin/plugin.json
+++ b/plugins/glean/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "glean-vnext",
   "displayName": "Glean vNext",
-  "version": "0.2.39",
+  "version": "0.2.40",
   "description": "Search and act across your company's apps — Jira, Slack, Salesforce, Google Workspace, and more — without leaving Cursor.",
   "author": {
     "name": "Glean"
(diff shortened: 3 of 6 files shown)


eshwar-sundar-glean:
this solution does not work on windows, tested it, it opens file explorer

eshwar-sundar-glean:
Pushed a fix and tested it on windows

eshwar-sundar-glean (approved):

PR #40, opened 2026-07-21
Cursor HITL: use Cursor's native approval (drop readOnlyHint + skip our elicitation)

pragati-agrawal-glean:
## Fix the Cursor HITL approval flow

**Symptom:** In Cursor, `requires_approval` actions via `run_tool` hang ~5 minutes with no approval prompt, then fail closed. Works fine in Claude Code.

## The fix — for Cursor, use Cursor's own native HITL instead of ours

Two paired changes, Cursor-only.

1. **Don't advertise `readOnlyHint` to Cursor** (`runToolAnnotations` now takes the client identity). Without the hint, Cursor shows its **native run-tool approval** before executing `run_tool`.
2. **Skip our elicitation gate for Cursor** (`!isCursorClient` added to the HITL gate in `handleRunTool`). With Cursor's native prompt already gating the call, our elicitation would be a redundant second gate — and it's the one that hangs. So Cursor executes directly after its native approval.

Demo video

https://github.com/user-attachments/assets/1d099748-2053-4597-bf8d-f8b853b972d5

Net: **Cursor → single native approval prompt; Claude Code → keeps `readOnlyHint` + our elicitation as before.**

## Validation
- `npm run typecheck` clean; `npx vitest run` **189/189**, incl. new cases: Cursor gets no `readOnlyHint`, and Cursor skips elicitation + executes directly.
- Before/after demo (video): two local servers built from the same codebase differing only by this PR — `main` build hangs on a write action in Cursor; PR build shows Cursor's native approval and completes.

## Notes / follow-ups
- Temporary until the Cursor-side elicitation-rendering regression is fixed upstream (bug report to file; repro available).
- `buildApprovalMessage`'s Cursor-specific branch is now unreachable (elicitation never fires for Cursor) — harmless; can be cleaned up separately.
- Latent hardening (not in this PR): the gate checks `getClientCapabilities()?.elicitation` truthy, but the SDK's `elicitInput` needs `.elicitation.form`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.40",
+  "version": "0.2.41",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.41",
+  "version": "0.2.42",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.41",
+  "version": "0.2.42",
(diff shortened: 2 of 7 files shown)


swarup-padhi-glean (approved):
Can we also have a demo video in the PR?

swarup-padhi-glean (approved):

pragati-agrawal-glean:
Demo video after making the change
https://github.com/user-attachments/assets/41fab32f-9dc0-45b9-ab2a-e68548e53e0f

swarup-padhi-glean:
Hey @pragati-agrawal-glean we will need to raise & investigate why we are having cursor's hitl for find_skills

eshwar-sundar-glean (approved):

PR #44, opened 2026-07-23
[Plugin] Fix intermittent re-auth from cross-process refresh-token rotation

pragati-agrawal-glean:
## Problem

The plugin can intermittently ask the user to authenticate again — `find_skills` or `run_tool` returns `[SETUP_REQUIRED]` — even though another live plugin process has already refreshed the shared OAuth grant.

## Root cause: cross-process refresh-token rotation

Each MCP session runs its own plugin process, but those processes share one credential file. The OAuth server rotates refresh tokens on every refresh and invalidates the presented refresh token. A sibling process can therefore retain a revoked refresh token in memory:

```text
process A: refreshes R0 → receives A1/R1 → persists A1/R1
process B: still holds R0 → refreshes → invalid_grant or invalid_request
```

Without coordination, the losing process can clear the shared store and surface `[SETUP_REQUIRED]`, potentially taking down the process that already has the valid rotated grant.

## Fix

1. **Treat the credential file as the source of truth.** `tokens()` reloads credentials from disk on every access through `syncTokensFromDisk()`. Correctness no longer depends on file mtime, and there is no mtime-based synchronization or environment override.
2. **Wait before clearing tokens.** `invalidateCredentials("tokens")` waits for a sibling's changed access token for a fixed two-second grace period, polling every 100 ms. If the sibling grant appears, it is adopted; only an unchanged store is cleared.
3. **Retry connect-level refresh failures.** `createRemoteClient()` observes whether a newer access token appeared on disk and retries the connection once. It also handles raw structured OAuth errors with `errorCode` `invalid_request` or `invalid_grant`, rather than matching human-readable error messages.
4. **Keep retries bounded and conservative.** A retry requires a changed token and is limited to one attempt, so unrelated OAuth errors and repeated failures are not hidden or retried indefinitely.
5. **Preserve atomic, private credential writes.** Credential updates continue to use temp-file-plus-rename writes, with `chmodSync(tmpPath, FILE_MODE)` applied before the rename so temporary files are `0600` as well.

The registered OAuth client is not replaced during a sibling retry; the new MCP connection reuses the same provider and the newly persisted grant. DCR client reuse remains covered separately by #45.

## Verification

Against the non-production SST endpoint:

```text
https://salessavvy-test-be.glean.com/mcp/gateway/proxy
```

with `client_api.oauth.accessToken.defaultExpiryMinutes=2`:

- Two separate Claude Code processes were launched with `--strict-mcp-config`.
- Both processes used the candidate plugin MCP server and one shared test credential file.
- A newly issued access token reported `expires_in=120` and the credential file remained mode `0600`.
- After access-token expiry, both real MCP calls succeeded without `[SETUP_REQUIRED]`.
- Local fingerprints confirmed that both the access token and rotated refresh token changed, while the registered client ID stayed stable.
- No raw credentials are included here.

## Files

- `src/auth-provider.ts`: direct disk synchronization, fixed grace-window sibling adoption.
- `src/remote-client.ts`: structured OAuth-error handling and bounded sibling retry.
- `src/token-store.ts`: atomic credential persistence and temp-file permission hardening.
- Corresponding auth-provider, remote-client, and token-store tests.

— sent via Glean Desktop

### DIFF
--- a/src/auth-provider.ts
+++ b/src/auth-provider.ts
@@ -7,7 +7,12 @@ import type {
 import { execFile, spawn } from "node:child_process";
 import { platform } from "node:os";
 import { getCallbackUrl, setExpectedState } from "./auth-callback-server.js";
-import { clearCredentials, loadCredentials, saveCredentials } from "./token-store.js";
+import {
+  clearCredentials,
+  credentialsMtimeMs,
+  loadCredentials,
+  saveCredentials,
+} from "./token-store.js";
 
 export type InvalidationScope = "all" | "client" | "tokens" | "verifier";
 
@@ -47,6 +52,10 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
   // explicitly invalidating. Used to detect when a previous auth URL didn't
   // complete — likely because the server rejected the (stale) client_id.
   private _authUrlPending = false;
+  // mtime (epoch ms) of the credentials file at the point our in-memory
+  // _tokens/_clientInfo snapshot was last taken from disk. Undefined until the
+  // first successful read. Used to detect sibling rewrites — see syncFromDisk.
+  private _credentialsMtimeMs: number | undefined;
 
   authorizationUrl: string | undefined;
 
@@ -63,6 +72,43 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
     if (stored) {
       this._tokens = stored.tokens as OAuthTokens | undefined;
       this._clientInfo = stored.clientInfo as OAuthClientInformationMixed | undefined;
+    }
+    this._credentialsMtimeMs = credentialsMtimeMs();
+  }
+
+  // Re-read the credentials file if it has changed on disk since we last read
+  // it. MCP servers are spawned per session, so several of our processes can be
+  // alive at once sharing one credentials file. When one process refreshes an
+  // (Ory-rotated, single-use) refresh token, it persists the new grant and
+  // silently invalidates the old refresh token that every other process still
+  // holds in memory. Without this, the next process to hit a 401 would refresh
+  // with its now-dead token, get invalid_grant, and force a full re-auth
+  // ([SETUP_REQUIRED]). Syncing here — right before the SDK reads tokens() at
+  // the start of its auth flow — lets us pick up the sibling's fresh grant
+  // instead. Guarded by mtime so the steady state is one stat(), not a re-parse.
+  //
+  // Conservative on removal: if the file is gone (undefined mtime) or has no
+  // tokens, we keep our in-memory snapshot rather than logging ourselves out —
+  // a transient stat failure or another process's explicit logout shouldn't
+  // drop a token that still works for us; a genuinely revoked token self-heals
+  // through the normal 401 path.
+  private syncFromDisk(): void {
+    const mtimeMs = credentialsMtimeMs();
+    if (mtimeMs === undefined) return;
+    if (
+      this._credentialsMtimeMs !== undefined &&
+      mtimeMs <= this._credentialsMtimeMs
+    ) {
+      return;
+    }
+    const stored = loadCredentials();
+    this._credentialsMtimeMs = mtimeMs;
+    if (!stored) return;
+    if (stored.tokens) {
+      this._tokens = stored.tokens as OAuthTokens;
+    }
+    if (stored.clientInfo) {
+      this._clientInfo = stored.clientInfo as OAuthClientInformationMixed;
     }
   }
 
@@ -84,9 +130,11 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
   saveClientInformation(info: OAuthClientInformationMixed): void {
     this._clientInfo = info;
     saveCredentials(this._tokens, this._clientInfo);
+    this._credentialsMtimeMs = credentialsMtimeMs();
   }
 
   tokens(): OAuthTokens | undefined {
+    this.syncFromDisk();
     return this._tokens;
   }
 
@@ -94,6 +142,9 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
     this._tokens = tokens;
     this._authUrlPending = false;
     saveCredentials(this._tokens, this._clientInfo);
+    // Record the mtime of the file we just wrote so our own write doesn't look
+    // like a sibling change on the next syncFromDisk.
+    this._credentialsMtimeMs = credentialsMtimeMs();
     this.onTokensChanged?.(tokens);
   }
 
--- a/src/auth-provider.ts
+++ b/src/auth-provider.ts
@@ -5,6 +5,7 @@ import type {
   OAuthTokens,
 } from "@modelcontextprotocol/sdk/shared/auth.js";
 import { execFile, spawn } from "node:child_process";
+import { setTimeout as sleep } from "node:timers/promises";
 import { platform } from "node:os";
 import { getCallbackUrl, setExpectedState } from "./auth-callback-server.js";
 import {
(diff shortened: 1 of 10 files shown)


src/auth-provider.ts
```diff
@@ -64,6 +73,43 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
...
+  // ([SETUP_REQUIRED]). Syncing here — right before the SDK reads tokens() at
+  // the start of its auth flow — lets us pick up the sibling's fresh grant
+  // instead. Guarded by mtime so the steady state is one stat(), not a re-parse.
+  //
+  // Conservative on removal: if the file is gone (undefined mtime) or has no
+  // tokens, we keep our in-memory snapshot rather than logging ourselves out —
+  // a transient stat failure or another process's explicit logout shouldn't
+  // drop a token that still works for us; a genuinely revoked token self-heals
+  // through the normal 401 path.
+  private syncFromDisk(): void {
```
eshwar-sundar-glean: #nit
syncTokensFromDisk?

src/auth-provider.ts
```diff
@@ -64,6 +73,43 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
       this._tokens = stored.tokens as OAuthTokens | undefined;
       this._clientInfo = stored.clientInfo as OAuthClientInformationMixed | undefined;
     }
+    this._credentialsMtimeMs = credentialsMtimeMs();
+  }
+
+  // Re-read the credentials file if it has changed on disk since we last read
+  // it. MCP servers are spawned per session, so several of our processes can be
+  // alive at once sharing one credentials file. When one process refreshes an
```
eshwar-sundar-glean: did we figure why this would even happen
```
When one process refreshes an
  // (Ory-rotated, single-use) refresh token, it persists the new grant and
  // silently invalidates the old refresh token that every other process still
  // holds in memory. Without this, the next process to hit a 401 would refr
```

swarup-padhi-glean: @eshwar-sundar-glean QQ:
Is the question why this will happen as to why does the refresh token get invalidated? - isn't it just because they are stale? Or is the question something else?

pragati-agrawal-glean:
Before changes:
Screenshot 2026-08-03 at 1 10 55 AM

After changes:
Screenshot 2026-08-03 at 1 12 43 AM

src/auth-provider.ts
pragati-agrawal-glean: Yes — this is a cross-process refresh-token rotation race. The host can keep multiple MCP server processes alive, and each provider initially snapshots the shared credentials file in memory. When one process refreshes, Ory rotates the single-use refresh token and persists the new grant, so another process can still hold the now-invalid old refresh token. Its next 401 then becomes invalid_grant and the SDK falls into full re-auth ([SETUP_REQUIRED]). The disk sync, invalidation adoption, and bounded reconnect retry cover both timing windows; the added tests exercise each case.

— sent via Glean Pi

src/token-store.ts
```diff
@@ -28,18 +28,32 @@ export function loadCredentials(): StoredCredentials | undefined {
...
     fs.chmodSync(dir, DIR_MODE);
     const data: StoredCredentials = { tokens, clientInfo };
-    fs.writeFileSync(filePath, JSON.stringify(data, null, 2), {
+    // Temp-file + rename: concurrent readers never see a half-written store.
+    const tmpPath = `${filePath}.${process.pid}.tmp`;
+    fs.writeFileSync(tmpPath, JSON.stringify(data, null, 2), {
       encoding: "utf-8",
       mode: FILE_MODE,
     });
-    fs.chmodSync(filePath, FILE_MODE);
```
mohit-gupta-glean, line 38: #clarify why is chmodsync call not needed anymore?

src/auth-provider.ts
```diff
@@ -64,6 +88,87 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
       this._tokens = stored.tokens as OAuthTokens | undefined;
       this._clientInfo = stored.clientInfo as OAuthClientInformationMixed | undefined;
     }
+    this._credentialsMtimeMs = credentialsMtimeMs();
```
mohit-gupta-glean: #suggest can we get the MTime also from loadCredentials instead of two separate file reads?

src/auth-provider.ts
```diff
@@ -64,6 +88,87 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
...
+  }
+
+  // Re-read the shared store after a sibling process rewrites it, so we use
+  // the rotated grant instead of a stale in-memory copy.
+  private syncTokensFromDisk(): void {
```
mohit-gupta-glean, line 79: Reading files multiple times in one flow is prone to race conditions. It will be better if we could have a combined function that gives both credentials and mtime.

mohit-gupta-glean: https://share.google/aimode/TCqFJXehUgfvJNBu9

src/auth-provider.ts
```diff
@@ -64,6 +88,87 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
...
+      this._tokens = stored.tokens as OAuthTokens;
+    }
+    if (stored.clientInfo) {
+      this._clientInfo = stored.clientInfo as OAuthClientInformationMixed;
+    }
+  }
+
+  // On invalid_grant, adopt a sibling's newer on-disk token instead of
+  // clearing. Returns false when nothing newer exists.
+  private adoptNewerTokenFromDisk(): boolean {
```
mohit-gupta-glean: This looks exactly the same as syncTokensFromDisk to me.

src/auth-provider.ts
```diff
@@ -64,6 +88,87 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
...
+  }
+
+  // Poll briefly for the race winner's write before clearing. Skipped when
+  // no refresh token was held (no race possible).
+  private async adoptNewerTokenWithGrace(): Promise<boolean> {
+    if (this.adoptNewerTokenFromDisk()) return true;
+    if (!this._tokens?.refresh_token) return false;
+    const deadline = Date.now() + rotationGraceMs();
+    while (Date.now() < deadline) {
+      await sleep(ROTATION_POLL_MS);
```
mohit-gupta-glean: why do we need to define sleep. It must be a standard function?

src/auth-provider.ts
```diff
@@ -64,6 +88,87 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
...
+      if (this.adoptNewerTokenFromDisk()) return true;
+    }
+    return false;
+  }
+
+  // Grace-bounded wait for a sibling's refresh; covers failures the SDK does
+  // not route through invalidateCredentials (e.g. invalid_request collisions).
+  async waitForSiblingRefresh(
```
mohit-gupta-glean, line 92: This again seems very similar to adoptNewerTokenWithGrace

src/auth-provider.ts
```diff
@@ -84,16 +189,20 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
...
 
   tokens(): OAuthTokens | undefined {
+    this.syncTokensFromDisk();
     return this._tokens;
   }
 
   saveTokens(tokens: OAuthTokens): void {
     this._tokens = tokens;
     this._authUrlPending = false;
     saveCredentials(this._tokens, this._clientInfo);
```
mohit-gupta-glean, line 132: would prefer to get the Mtime from Save call itself. otherwise, there is again a race condition if two sessions write close to each other. It is possible that a sibling write gets lost as we read the mtime separately after writing.

src/auth-provider.ts
```diff
@@ -113,6 +222,8 @@ export class GleanOAuthClientProvider implements OAuthClientProvider {
         saveCredentials(this._tokens, undefined);
         break;
       case "tokens":
+        // Usually a sibling's rotation — try adopting before clearing.
+        if (await this.adoptNewerTokenWithGrace()) return;
```
mohit-gupta-glean: We seem to be calling invalidateCredentials only with all scope. Which path does this change impact?

src/auth-provider.ts
pragati-agrawal-glean: Done

pragati-agrawal-glean:
Simplified the implementation - removed the modified time based logic and added logic to always read the disk stored tokens before using them to ensure a process always has the latest tokens (This assumes new tokens will be written to disk within 2 seconds and the disk write will execute successfully)

src/auth-provider.ts
pragati-agrawal-glean: simplified the implementation to have just one function for this. Thanks

src/auth-provider.ts
pragati-agrawal-glean: mtime based implementation removed

src/token-store.ts
pragati-agrawal-glean: when a new temp file is created to write to it, it should have the permission 0o600. Still adding the chmodSync for safety. Thanks for pointing out.

src/auth-provider.ts
pragati-agrawal-glean: in the simplified implementation now there is only one read.

PR #46, opened 2026-07-30
fix: security review hardening (approval gate, hook scope, oauth, refresh, args)

eshwar-sundar-glean:
## Summary

Ports the security review fixes from gleanwork/agent-plugins#6 (from @swarup-padhi-glean's review) to this source repo, since the `glean-vnext` server code originates here.

Adapted to this repo's shape: the local server is named **`glean`** (no coexist/`glean-local` split), the discovery tool is still **`find_skills`**, and CI (`ci.yml`) already runs typecheck + tests, so no workflow change is needed.

## Fixes

| Sev | Area | Fix |
|-----|------|-----|
| **High** | `src/tools/run-tool.ts` | Approval gate **fails closed**: a missing/unparseable tool JSON no longer bypasses the HITL gate while the native prompt is suppressed (`readOnlyHint`). Only skip when `requires_approval` is explicitly `false`. |
| **Med** | `src/auth-provider.ts` | OAuth `state` CSRF check made **live** — generate a `state` nonce on the authorize URL so the loopback callback's check runs (was dead code). |
| **Med** | `src/remote-client.ts` | **Serialize** `connect`/`finishAuth` so concurrent handlers don't race the SDK token refresh and trip `invalidateCredentials` on refresh-token rotation. |
| **Med** | `src/tools/approval-args.ts` | **Sanitize** approval-prompt argument keys (collapse newlines) + disclose omitted args with `(+N more)`. |
| **Med** | `src/skill-writer.ts` | Match `SKILL.md` under either path separator (Windows `\`). |

> The auto-approve **hook matcher scoping** change was reverted from this PR and deferred to a separate change/discussion.

## Testing

- `npm run typecheck` — clean
- `npm test` — **192/192** (3 new: fail-closed accept/decline, arg-key injection)
- `npm run build` — `dist/` rebuilt and committed
- Version bumped **0.2.42 → 0.2.43** across all three host manifests

## Companion PR
gleanwork/agent-plugins#6 (same fixes in the consolidated repo).

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.42",
+  "version": "0.2.43",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.42",
+  "version": "0.2.43",
   "description": "Glean Codex plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.cursor-plugin/plugin.json
+++ b/plugins/glean/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "glean-vnext",
   "displayName": "Glean vNext",
-  "version": "0.2.42",
+  "version": "0.2.43",
   "description": "Search and act across your company's apps — Jira, Slack, Salesforce, Google Workspace, and more — without leaving Cursor.",
   "author": {
     "name": "Glean"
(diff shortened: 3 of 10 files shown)


plugins/glean/hooks/hooks.json
```diff
@@ -2,7 +2,7 @@
   "hooks": {
     "PreToolUse": [
       {
-        "matcher": "mcp__.*glean.*run_tool",
```
eshwar-sundar-glean, line 5: there seems to be some issues with this sometimes. @swarup-padhi-glean it might be very tricky to maintain a very exact match. I am planning on reverting this to previous state.

plugins/glean/hooks/auto-approve-run-tool.mjs
```diff
@@ -31,10 +31,16 @@ try {
 }
 
 const toolName = String(input.tool_name ?? "");
-const bareName = toolName.split("__").pop() ?? "";
-// Scope strictly to this plugin's run_tool — the tool name carries the glean
-// plugin/server prefix (e.g. mcp__plugin_glean-vnext_glean__run_tool).
-if (!toolName.includes("glean") || bareName !== "run_tool") {
+// Scope strictly to THIS plugin's server (glean). Matching any name that merely
```
eshwar-sundar-glean: The previous code was good enough, lets just remove this please

swarup-padhi-glean (approved):
Here can we test the HITL experience once before merge?

PR #49, opened 2026-08-18
Bake the plugin version into the bundle

mohit-gupta-glean:
## Problem

The shipped bundle contains no version of its own. On `main`:

```
$ grep -c "0.2.43" plugins/glean/dist/index.js
0
```

Meanwhile both MCP constructors report a hardcoded `"1.0.0"` that has never matched a release — `src/index.ts:129` and `src/remote-client.ts:188`. The real version lives only in the three host manifests.

So today the plugin cannot tell a remote which version is running, and the version it *does* report is wrong.

## Change

`scripts/build.mjs` reads the version from `.claude-plugin/plugin.json` and substitutes it via esbuild `define` into a new `src/version.ts`. Both constructors now report it, so the version is visible on the MCP handshake in each direction.

After this change:

```
$ grep -c "0.2.44" plugins/glean/dist/index.js
1
```

## Why a build constant rather than reading a manifest at runtime

This value is what a remote policy would use to decide whether a plugin is outdated, blocked, or unsupported, so it has to describe the code that is actually running.

A manifest sits beside the bundle as editable JSON, so a runtime read reports only what the plugin *claims*. Worse, a *fallback* to one would convert "the constant is missing" into "here is a version the user can edit" — a downgrade path for anyone able to make the constant unreadable.

`pluginVersion()` therefore reports `source: "build"` or `source: "unknown"` and never guesses. A dev run through `tsx` has no `define` applied and correctly reports `unknown`; callers are expected to skip version-dependent behaviour rather than treat the placeholder as real.

CI already verifies `dist/` is in sync with source, so the compiled literal cannot drift from the manifest unnoticed.

## The trade-off, and its mitigation

Baking it in makes the build step load-bearing: a silently missing constant would make every install report an unknown version and disable version-dependent behaviour fleet-wide — a wider blast radius than any single fallback.

So the build **fails** if the manifest version is absent or is not a plain `x.y.z`:

```
Error: build: plugins/glean/.claude-plugin/plugin.json must declare a plain x.y.z
version, got "not-a-version". The bundled constant is what the plugin reports about
itself, so an absent or malformed value fails the build rather than shipping.
```

Verified by temporarily corrupting the manifest.

## Testing

- `npm run typecheck` clean
- `npm test` — 192 passing (189 existing + 3 new)
- `npm run build` — dist committed and in sync
- version-bump gate: `0.2.43 → 0.2.44 (all manifests aligned)`
- confirmed the literal lands in `dist/index.js`, and that a malformed manifest version fails the build

The new tests deliberately exercise the `unknown` path, since vitest does not apply esbuild's `define` — the same condition as a `tsx` dev run.

## Context

This is a prerequisite for remote capability/policy negotiation ([scio#273428](https://github.com/askscio/scio/pull/273428)), where version rules can deactivate a plugin. It stands on its own, though: reporting an accurate version on the MCP handshake is worth having regardless.

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -38,31 +38,58 @@ const nodeBuiltins = [
 // Baking it in makes this step load-bearing: a silently absent constant would make
 // every install report an unknown version and disable any version-dependent
 // behaviour fleet-wide, which is a far worse outcome than a red build.
-const MANIFEST = "plugins/glean/.claude-plugin/plugin.json";
+//
+// Every host manifest is read, not just one. Reading a single manifest would bake in
+// whichever host happened to be picked and silently disagree with the others if they
+// ever drifted. scripts/check-version-bump.sh enforces that all three match, but only
+// in CI — a local build has to catch it too, since that is where the constant is made.
+const MANIFESTS = [
+  "plugins/glean/.claude-plugin/plugin.json",
+  "plugins/glean/.codex-plugin/plugin.json",
+  "plugins/glean/.cursor-plugin/plugin.json",
+];
+
+// Same pattern as SEMVER_RE in scripts/check-version-bump.sh, and deliberately stricter
+// than \d+: that would accept a non-canonical "01.02.003", which CI then rejects. The
+// two checks must agree, or a local build can produce a version CI refuses.
+const SEMVER_RE = /^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$/;
 
 function pluginVersionFromManifest() {
-  let raw;
-  try {
-    raw = JSON.parse(readFileSync(MANIFEST, "utf-8"));
-  } catch (err) {
-    throw new Error(`build: cannot read ${MANIFEST}: ${err.message}`);
-  }
-  const version = raw.version;
-  if (typeof version !== "string" || !/^\d+\.\d+\.\d+$/.test(version)) {
+  const found = MANIFESTS.map((manifest) => {
+    let raw;
+    try {
+      raw = JSON.parse(readFileSync(manifest, "utf-8"));
+    } catch (err) {
+      throw new Error(`build: cannot read ${manifest}: ${err.message}`);
+    }
+    const version = raw.version;
+    if (typeof version !== "string" || !SEMVER_RE.test(version)) {
+      throw new Error(
+        `build: ${manifest} must declare a plain x.y.z version, got ${JSON.stringify(version)}. ` +
+          `The bundled constant is what the plugin reports about itself, so an absent or ` +
+          `malformed value fails the build rather than shipping.`,
+      );
+    }
+    return { manifest, version };
+  });
+
+  const versions = [...new Set(found.map((f) => f.version))];
+  if (versions.length > 1) {
     throw new Error(
-      `build: ${MANIFEST} must declare a plain x.y.z version, got ${JSON.stringify(version)}. ` +
-        `The bundled constant is what the plugin reports about itself, so an absent or ` +
-        `malformed value fails the build rather than shipping.`,
+      `build: plugin manifests disagree on the version, so there is no single value to ` +
+        `bake in:\n` +
+        found.map((f) => `  ${f.version}  ${f.manifest}`).join("\n"),
     );
   }
-  return version;
+  return versions[0];
 }
 
 const pluginVersion = pluginVersionFromManifest();
+const OUTFILE = "plugins/glean/dist/index.js";
 
 await build({
   entryPoints: ["src/index.ts"],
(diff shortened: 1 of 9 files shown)


scripts/build.mjs
```diff
@@ -18,19 +18,59 @@
...
+// to be the version of the code that is actually running. A manifest read at runtime
+// reports only what an editable file next to the bundle claims. CI already verifies
+// that `dist/` is in sync with source, so a compiled-in literal cannot drift from the
+// manifest unnoticed.
+//
+// The build FAILS rather than emitting a bundle with a missing or malformed version.
+// Baking it in makes this step load-bearing: a silently absent constant would make
+// every install report an unknown version and disable any version-dependent
+// behaviour fleet-wide, which is a far worse outcome than a red build.
+const MANIFEST = "plugins/glean/.claude-plugin/plugin.json";
```
eshwar-sundar-glean: Assumes all plugins versions will be bumped together and uses claude? LGTM, cant think of corner cases

src/version.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * The version this build reports, plus whether it is trustworthy.
+ *
+ * `source: "build"` in every shipped release; `source: "unknown"` only for a dev run
+ * or a broken build, where the placeholder version must not be treated as real.
+ */
+export function pluginVersion(): ResolvedVersion {
+  if (BUILD_VERSION) return { version: BUILD_VERSION, source: "build" };
+  // A placeholder rather than an empty string, so callers never have to handle a
```
eshwar-sundar-glean, line 50: We are throwing at build, will this ever happen?

eshwar-sundar-glean (approved):
LGTM

scripts/build.mjs
mohit-gupta-glean: Good catch — it did assume that, and I have removed the assumption rather than documented it.

`check-version-bump.sh` does enforce that all three manifests share one version, but only in CI. A local `npm run build` with drifted manifests would have silently baked in whichever host was named here, which is exactly the corner case you were reaching for.

`build.mjs` now reads all three, validates each, and fails if they disagree:

```
Error: build: plugin manifests disagree on the version, so there is no single value to bake in:
  0.2.44  plugins/glean/.claude-plugin/plugin.json
  9.9.9   plugins/glean/.codex-plugin/plugin.json
  0.2.44  plugins/glean/.cursor-plugin/plugin.json
```

So no host is privileged and nothing depends on the bump discipline holding. Verified by drifting the codex manifest by hand.

src/version.ts
mohit-gupta-glean: Yes, but never in a shipped bundle — and my comment was misleading about why, which is what prompted the question. It said `unknown` covers "a dev run or a broken build", and the second half is wrong: the build cannot produce a bundle without the constant.

The branch is reached when the code runs **without having been bundled at all** — `npm run dev` and any other tsx entry, where esbuild never applies the define. `typeof __GLEAN_PLUGIN_VERSION__` is then `"undefined"`, so `unknown`, so a dev run enforces no version policy. That is deliberate: a dev build has no meaningful version to enforce against.

Reworded to say that instead.

Your question also exposed a real gap. The build validated its *input* (the manifest) and never checked its *output*, so if the define ever stopped applying — identifier renamed on one side only, define key lost in a merge — the build would succeed, `dist/` would still match a fresh build so CI stays green, and every install would silently report `unknown` with version enforcement off fleet-wide. The build now reads the bundle back and fails if the placeholder survived or the literal is missing:

```
Error: build: plugins/glean/dist/index.js still contains the __GLEAN_PLUGIN_VERSION__ placeholder,
so the esbuild define did not substitute it.
```

Verified by renaming the define key. CI runs `npm run build`, so it is covered without a separate job.

scripts/build.mjs
```diff
@@ -18,19 +18,59 @@
...
+
+function pluginVersionFromManifest() {
+  let raw;
+  try {
+    raw = JSON.parse(readFileSync(MANIFEST, "utf-8"));
+  } catch (err) {
+    throw new Error(`build: cannot read ${MANIFEST}: ${err.message}`);
+  }
+  const version = raw.version;
+  if (typeof version !== "string" || !/^\d+\.\d+\.\d+$/.test(version)) {
```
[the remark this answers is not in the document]
mohit-gupta-glean: Fixed — this was a real inconsistency. `build.mjs` now uses the same pattern as `SEMVER_RE` in `scripts/check-version-bump.sh`, so a local build can no longer produce a version CI rejects:

```js
const SEMVER_RE = /^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$/;
```

Verified against `01.02.003`, which now fails the build rather than being baked in.

PR #50, opened 2026-08-18
Capability and policy negotiation: report context, record policy

mohit-gupta-glean:
## What

Adds capability & policy negotiation to the plugin: it reports its context to the remote in a `_meta` vendor extension, and records whatever policy comes back.

**It enforces nothing.** No tool is dropped, no call is refused, `file_args` stays in the schema, HITL is untouched. A policy saying `deactivated` would be computed, logged, cached — and ignored. Nothing outside `src/policy/` reads the resolved decision.

That is *not* the same as a no-op. Five effects are observable:

| Effect | Where |
|---|---|
| New `_meta` on every remote request (`tools/list`, `tools/call`) | `remote-client.ts`, `remote-passthrough.ts` |
| Writes a new file, `~/.glean/policy-cache.json` | `policy/cache.ts` |
| New log lines: `policy.resolved`, `policy.malformed`, `policy.unknown-keys` | `policy/session.ts` |
| Wraps the stdio transport to observe the negotiated MCP revision | `index.ts` |
| `setup` output gains a policy summary | `index.ts` |

Plus, via #49: `serverInfo`/`clientInfo` version changes from a hardcoded `"1.0.0"` to the real build constant.

**Stacked on #49** — base is `mohit/plugin-version-constant`. Review #49 first.

## Verified against production

Installed to `~/.cursor/plugins/local/glean` and exercised in Cursor against `scio-prod-be.glean.com`. Every path that carries `_meta` was confirmed accepted by the real gateway:

```
policy.resolved {"label":"tools/list","outcome":"no-policy",...}
policy.resolved {"label":"tools/call(find_skills)","outcome":"no-policy",...}
policy.resolved {"label":"tools/call(run_tool)","outcome":"no-policy",...}
policy.resolved {"label":"tools/call(search)","outcome":"no-policy",...}
```

`tools/call(search)` is the promoted-passthrough path, `run_tool` exercised the full find_skills → skill → downstream flow. The `setup` output confirmed each field's provenance:

```
Plugin version: 0.2.45 (source: build)
Host: cursor-vscode 1.0.0 (source: handshake)
MCP revision: 2025-11-25
Server inventory: unavailable
Policy: version unenforced, features {toolPromotion,metaTools,hitl,fileArgs all true}
```

So: the build constant survives into a real install (#49's premise), host identity comes from the peer's `clientInfo` rather than our own `serverInfo`, the transport shim observes the negotiated revision, and the deferred inventory reports `unavailable` rather than implying an empty list.

`policy-cache.json` was **never created**, which is correct — only a *valid* policy writes it, and production returns none. `no-policy` stayed distinct from unreachable throughout.

## What testing changed

Three commits exist only because the live test exposed the problem:

- **Negotiation was effectively unobservable.** Against a remote with no policy support every response resolves to the same decision, so change-only logging went silent after the first exchange and never reported `tools/call` again. Confirming the mechanism ran meant reading code, not logs. Now logs on `changed || first-time-for-this-label` — bounded to under ten lines per process.
- **No pid in the log.** Two identical `tools/list` lines a second apart took a `pgrep` and three discarded theories to attribute. Every line now carries `[pid]`; the interesting state (in-force decision, fallback session id, log dedup) is all per-process, so lines that cannot be attributed to a process cannot explain it.
- **Cursor reuses one process across chat sessions** — verified by invoking `setup` in a third session and seeing the same pid serve it. Multiple concurrent processes are therefore not routine, which downgrades (but does not remove) the concern that two processes could hold different in-force decisions once enforcement lands.

## Cache design

The policy cache and the pre-existing `remote-tools-cache-store.ts` are separate and operate independently:

- **The policy cache stores policy only.** The lifted module originally carried a `toolsList` half that duplicated the tools store — dead code shadowing a live subsystem.
- **Only the raw catalog is cached, never a policy-resolved surface.** `tools/list` composes its answer by filtering the catalog through the policy in force, at read time, so a catalog and a policy cached at different moments still yield the surface the current policy dictates. Recorded on `saveRemoteTools`, since that is where an optimisation would break it.
- **A cached policy is never discarded locally** — there is no clear function, and `setup({reset})` deliberately leaves it. It can carry a deactivation or version block, so any local discard is a way to shed one, and afterwards the remote may be unreachable with nothing to re-sync from. Re-keying covers instance switches: entries are keyed by remote URL.
- **Writes are atomic** (temp file + rename). `writeFileSync` truncates then writes, and the cache reads a parse failure as "no data" — so a process killed mid-write silently discarded every entry. Hosts do reap these processes, so that was a live path around the never-discard rule.

## Testing

`219 passed`. Includes the atomic-write helper, whose preservation test is verified by mutation: replacing `writeFileAtomicSync` with a plain `writeFileSync` fails exactly that test and no other.

### DIFF
--- a/plugins/glean/.claude-plugin/plugin.json
+++ b/plugins/glean/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.44",
+  "version": "0.2.45",
   "description": "Glean plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.codex-plugin/plugin.json
+++ b/plugins/glean/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "glean-vnext",
-  "version": "0.2.44",
+  "version": "0.2.45",
   "description": "Glean Codex plugin for discovering skills and running tools.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.cursor-plugin/plugin.json
+++ b/plugins/glean/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "glean-vnext",
   "displayName": "Glean vNext",
-  "version": "0.2.44",
+  "version": "0.2.45",
   "description": "Search and act across your company's apps \u2014 Jira, Slack, Salesforce, Google Workspace, and more \u2014 without leaving Cursor.",
   "author": {
     "name": "Glean"
--- a/plugins/glean/.cursor-plugin/plugin.json
+++ b/plugins/glean/.cursor-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "glean-vnext",
   "displayName": "Glean vNext",
   "version": "0.2.45",
-  "description": "Search and act across your company's apps \u2014 Jira, Slack, Salesforce, Google Workspace, and more \u2014 without leaving Cursor.",
+  "description": "Search and act across your company's apps — Jira, Slack, Salesforce, Google Workspace, and more — without leaving Cursor.",
   "author": {
     "name": "Glean"
   },
--- a/plugins/glean/dist/index.js
+++ b/plugins/glean/dist/index.js
@@ -25414,15 +25414,6 @@ function savePolicy(serverUrl, policy) {
   };
   writeAll(all);
 }
-function clearCache(serverUrl) {
-  if (!serverUrl) {
-    writeAll({});
-    return;
-  }
-  const all = readAll();
-  delete all[serverUrl];
-  writeAll(all);
-}
 
 // src/policy/protocol-version.ts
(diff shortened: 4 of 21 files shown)


mohit-gupta-glean:
**Note on CI:** the workflow is scoped to `pull_request: branches: [main]`, and this PR targets `mohit/plugin-version-constant` rather than `main`, so GitHub reports no checks here. That is the workflow config, not a failure.

Once #49 merges, retarget this PR to `main` and CI will run. In the meantime the same checks were run locally against this branch:

| Check | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm test` | 207 passing |
| `npm run build` | succeeds |
| `git diff --exit-code plugins/glean/dist/` | clean — dist in sync |
| `scripts/check-version-bump.sh origin/main` | `0.2.43 → 0.2.45` |

eshwar-sundar-glean (approved):

